PR #16, opened 2018-03-30
WIP: Use agent for timeseries streams

intrepidlemon:


blackfynn/api/timeseries.py
```diff
@@ -388,6 +265,8 @@ def get_ts_data_iter(self, ts, start, end, channels, chunk_size,
         the_start = ts.start if start is None else infer_epoch(start)
 
         # chunk
+        if chunk_size is None:
+            chunk_size = 100
```
merky: Is this for testing, or do we need to handle the chunk_size argument better?

intrepidlemon: I think the old api allowed chunk_size to be null but the agent complains. so it seems like our options are: 
- make agent complain less
- set the default here

scirner22: What was the behavior of the old api when chunk size was none?

merky: Chunk size is a somewhat python-only thing. The API requests were just for some span of time, but the chunk size here dictated what span of times were requested. I don't believe an API call received a `chunk` argument at all, only `start`, `end` — but I may be wrong.

intrepidlemon: ah, it seems like the python client used to re-partition based on chunk size here: https://github.com/Blackfynn/blackfynn-python/pull/16/files#diff-c29515fbacbc2950e6b2937b08b73108L161

I couldn't quite understand what happens when chunk_size is None though. It doesn't seem like it should succeed in that case.

merky:
@scirner22 Correct, we should be returning an empty Dataframe (or Series) if a chunk is empty.

@intrepidlemon See my  bug fix here which addresses a current chunking issue Steve discovered (but also returns an empty Series when there is no data within a valid iterator chunk). https://github.com/Blackfynn/blackfynn-python/pull/18

intrepidlemon:
@merky this PR largely removes that section of code since it depends on the agent to return valid chunks and so I don't need to adapt your fix right? 
As this PR is currently implemented, it will return empty DataFrames for empty chunks.

merky:
@intrepidlemon I think you're good-to-go. Like you said, my fix was for the previous paging/request logic, so you should be good as long as you yield empty Dataframes on empty chunks 👍

PR #18, opened 2018-04-12
Bug fix: Retrieve chunk slice based on time index

merky:
When retrieving time series data in an iterative/chunked manner, e.g. using `.get_data_iter`, there is a possibility of incorrect chunk responses when gaps are present in the data. 

Note: this is a hot-fix for the retrieval mechanism in the existing client. In the near future, all time series requests are rerouted through the agent (pending release).


merky:
@scirner22 This should resolve any issues you've encountered before. Let me know if any discrepancies still persist, otherwise I'll apply this hot-fix to the current client.

PR #21, opened 2018-04-23
Upload using preview

intrepidlemon:
- [x] handle warnings


blackfynn/api/transfers.py
```diff
@@ -279,17 +279,47 @@ def upload_files(self, destination, files, dataset=None, append=False, display_p
...
+            endpoint = self._uri('/files/upload/preview'),
+            params = params,
+            json=payload,
+            )
+
+        import_id_map = dict()
+        for p in response.get("packages", list()):
+            import_id = p.get("importId")
+            # TODO: display warnings here somehow
+            warnings = p.get("warnings", list())
```
merky: For now, maybe use `logger.warn` to spit these out?

blackfynn/api/transfers.py
```diff
@@ -279,17 +279,47 @@ def upload_files(self, destination, files, dataset=None, append=False, display_p
                     raise future.exception()
 
         # trigger ETL import
-        return self.set_upload_complete(group_id, dataset_id, destination_id, append)
+        return [self.set_upload_complete(import_id, dataset_id, destination_id, append) for import_id in set(import_id_map.values())]
```
merky: Is it correct to say that `self.set_upload_complete` will be called after *all* files are done uploading? It seems that way. I think it may be more beneficial to execute this per group in `import_id_map`, such that you call the "complete" endpoint after all the files are finished uploading *per package* — this enhances user experience, as file sources for some packages may be quick to upload compared to others, and you'd want to start/finish their processing (on the server side) as soon as you can. This could be accomplished by inserting some logic in the the `as_completed` loop above. Let me know if you have any questions about this.

blackfynn/cli/bf.py
```diff
@@ -15,14 +15,11 @@
   props             Add/remove/modify a package or collection's properties
   move              Move a package or collection
   where             Show path to package or collection
-  append            Append data to a package
   rename            Rename a package or collection
   delete            Delete a package or collection
   create            Create a collection
   get               Get the contents of a package, collection, or dataset
 
-  upload            Upload file(s) or directory
```
merky: Is the deletion of these CLI commands assuming availability of the rust-agent? I wonder if we'd release this PR before agent is packaged with python client -- thoughts?

intrepidlemon: These changes do assume that the agent will be available, but I can add them back if not the case. 

@scirner22 @mikeswoods have we deciding on anything about when/how the client will be packaged with the agent? 

If we release this before the agent is available, I can go ahead and add back all of this upload logic.

scirner22: I believe it will be later this month. It's probably best to rip this stuff out in a separate PR.

intrepidlemon:
@merky comments addressed

merky:
:shipit: !

PR #22, opened 2018-05-09
WIP: modularize application

intrepidlemon:


blackfynn/extensions.py
```diff
@@ -0,0 +1,38 @@
+import sys
+
+
+# try to import optional packages
+try:
+    import pandas
+except ImportError:
+    pd = None
```
merky: This would have to be `pandas`, I believe.

blackfynn/api/timeseries.py
```diff
@@ -502,6 +503,8 @@ def get_annotation(self, ts, layer, annot):
         return TimeSeriesAnnotation.from_dict(resp["annotation"], api=self.session)
 
     def iter_annotations(self, ts, layer, window_size=10, channels=None):
+        check_extensions()
```
merky: I don't believe annotations deal with numpy/pandas at all, so are probably safe methods in "light" mode.

intrepidlemon: annotations uses the numpy method `ceil` but I could probably reimplement that without numpy

https://github.com/Blackfynn/blackfynn-python/pull/22/files/b100ce68923de55adcc84fc3787b95de24786999#diff-c29515fbacbc2950e6b2937b08b73108L525

PR #25, opened 2018-06-08
Rename concepts

merky:
Changes:
- Renames `Concept` to `Model` and `ConceptInstance` to `Record`
- Adjusts user-facing method names to match
- Adds iterator to `Model` so you can do `for each item in model`

**NOTE:** Code  in `blackfynn.api.concepts` has not been modified, as the REST API still uses `/concepts`-based terminology.

Example: 

```python
bf = Blackfynn()
ds = bf.get_dataset('hetnet.io')

# create model
disease = ds.create_model('disease', 'Disease')
disease.add_property('name', str)
disease.add_property('category', str)

# create record(s)
disease.create_record({
  'name': 'Parkinsons',
  'category': 'Neurological'
})
```


blackfynn/api/concepts.py
```diff
@@ -32,72 +32,72 @@ def _get_relationship_type(self, relationship, instance = None):
...
             r['schema'] = self.update_properties(dataset, concept)
-        return Concept.from_dict(r, api=self.session)
+        return Model.from_dict(r, api=self.session)
 
     def create(self, dataset, concept):
-        assert isinstance(concept, Concept), "concept must be type Concept"
+        assert isinstance(concept, Model), "concept must be type Model"
         dataset_id = self._get_id(dataset)
         r = self._post(self._uri('/{dataset_id}/concepts', dataset_id=dataset_id), json=concept.as_dict())
         concept.id = r['id']
```
jwagenaar: keeping these like this?

jwagenaar: "concept must be type Model" seems odd feedback to the user

PR #28, opened 2018-06-19
New `type` property added to Datasets

sLopezdeDiego:


merky:
Ignoring Dataset `type` field for now, will address in future; conflicts with base class `.type` property.

PR #30, opened 2018-06-28
Revisit graph methods

merky:
### NEW:
- Added `model.remove_property(...)` method
- Added Record API endpoints for 
  - `get_all_related`
  - `get_all_related_of_type`
  - `get_counts`

### REMOVED:
The following methods caused confusion, and generally, wanted to have graph methods be more record-oriented, and avoid treating relationships as objects.
- `record.neighbors()`
- `record.links()`
- `record.relationships()`

( see `record.get_related()` which accomplish above)

### CHANGES:
- `record.link(...)` is now `record.relate_to(...)`
- Use `record.get_related()` to get any related records
   - Removed `record.neighbors()`, `record.links()`, and `record.relationships()` as they caused confusion, and generally focused on being more record-oriented, and avoided treating relationships as objects.
- `record.files` is now `record.get_files()`
- `dataset.create_relationship(...)` is now `dataset.create_relationship_type(...)`
- `model.create_records(v1, v2, ...)` is now `model.create_records([v1, v2, ...])` (not using `*args` format).
- `relationship_type. create_many(...)` is now `relationship_type.create(...)`
- Updated/added to docstrings for better documentation.

### DIFF
--- a/blackfynn/models.py
+++ b/blackfynn/models.py
@@ -2614,7 +2614,7 @@ class Model(BaseModelNode):
 
         Example::
 
-          mouse_002 = mouse.create({"id": 2, "weight": 2.2})
+          mouse_002 = mouse.create_record({"id": 2, "weight": 2.2})
 
         """
         self._check_exists()
@@ -2630,7 +2630,7 @@ class Model(BaseModelNode):
         ci = self._api.concepts.instances.create(self.dataset_id, ci)
         return ci
 
-    def create_records(self, *values_list):
+    def create_records(self, values_list):
         """
         Creates multiple records of the model on the platform.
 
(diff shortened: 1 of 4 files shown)


merky:
I'd love to hear your thoughts on the above changes!

blackfynn/models.py
```diff
@@ -2850,43 +2986,57 @@ def link(self, source, destination, values=dict()):
...
             assert value['relationship_type']==self.type, u'RelationshipType type of items need to match relationship type: "{}"'.format(self.type)
 
         li_list = [
             Relationship(
                 dataset_id  = self.dataset_id,
                 type        = item['relationship_type'],
-                source      = item['source'],
-                destination = item['destination'],
+                source      = item['from'],
+                destination = item['to'],
```
rohanshah: why use both `source` and `destination` and `to` and `from`?

merky: Motivation was to reduce input when specifying records in a list, `{'to': ..., 'from': ...}` vs {'source': ..., 'destination': ...}` You're absolutely right, though, as it does not match the interface to `Record`.

PR #33, opened 2018-07-24
Deploy library and docs via Travis CI

merky:
- Modified tests to work
  - Removed all tests dependent on a "super user"
  - Modified existing tests to work with current API
  - These are integration tests running against production API
  - Note: temporary dataset created/destroyed during testing
- Now using Travis CI
  - Runs tests on all branches
  - Deployment to S3 on tagged releases (`docs.blackfynn.io/python/<version>`)
- Flexible `requirements.txt` versioning
- Small bug fixes along the way (record->package linking)


requirements.txt
```diff
@@ -1,11 +1,11 @@
...
-boto3==1.4.4
-requests==2.18.3
-requests-futures==0.9.7
-numpy==1.13.1
-pandas==0.20.3
+boto3>=1.4
+requests>=2.18
+requests-futures>=0.9
+numpy>=1.13
+pandas>=0.20
```
zduey: awesome, thanks for doing this. I was going to suggest doing this if you didn't since it was causing some grief with graph-ingest that depends on the client

tests/test_timeseries.py
```diff
@@ -210,18 +206,18 @@ def test_timeseries_annotations(client, timeseries):
     #layer1.add_annotations([annot2,annot3])
     annot3 = timeseries.add_annotations(layer=layer1,annotations=annot3)
     assert annot3.exists
-    annot3.delete()
-    assert not annot3.exists
+    #annot3.delete()
```
dkh215: Why did we have to remove all the delete calls?

tests/test_concepts.py
```diff
@@ -4,7 +4,7 @@
 
 from blackfynn.models import Model, Record, DataPackage, RelationshipType, Relationship
 
-def test_concepts(dataset):
+def test_models(dataset):
```
dkh215: Just a general comment that we can address later, would split this function up similar to how other test suites are structured.

.travis.yml
```diff
@@ -0,0 +1,39 @@
+sudo: false
+language: python
+python:
+- '2.7'
+install:
+- pip install -r requirements.txt
+- pip install -r requirements-test.txt
```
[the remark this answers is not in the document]
merky: I think so, yes. It separates out dependencies for the library vs. development stuff. Our `setup.py` uses the contents of `requirements.txt`.

.travis.yml
```diff
@@ -0,0 +1,39 @@
...
+  - |
+    if [[ "$TRAVIS_BRANCH" == "master" ]]; then
+      make html-latest
+    fi
+  - cd ./_build/html
+deploy:
+  - provider: pypi
+    on:
+      tags: true
+    user: bfmark
```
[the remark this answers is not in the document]
merky: Yes, it does! Let me look into that...

requirements.txt
merky: There's actually a ticket for this somewhere, but I figured I would tackle it while I was in there.

tests/test_concepts.py
merky: Very true. The tests all need massaging... and there are missing tests, but this at leasts gets the existing tests working.

PR #34, opened 2018-07-26
APP-888: Better Error Handling For Invalid Model Properties

mattusifer:
Currently `dataset.create_model()` will fail if the given schema was invalid. This is bad because the model will have been created - just without properties - but to the user it looks as though the entire operation failed. 

This change is to log an error when the 'add properties' step fails, but successfully return the newly created model without properties.

### DIFF
--- a/blackfynn/api/concepts.py
+++ b/blackfynn/api/concepts.py
@@ -1,4 +1,7 @@
 # -*- coding: utf-8 -*-
+import logging
+import requests
+
 from blackfynn.api.base import APIBase
 from blackfynn.models import (
     Model, Record, RecordSet, ModelProperty, ProxyInstance,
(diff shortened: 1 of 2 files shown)


blackfynn/api/concepts.py
```diff
@@ -97,8 +100,21 @@ def create(self, dataset, concept):
         r = self._post(self._uri('/{dataset_id}/concepts', dataset_id=dataset_id), json=concept.as_dict())
         concept.id = r['id']
         r['dataset_id'] = r.get('dataset_id', dataset_id)
+
         if concept.schema:
-            r['schema'] = self.update_properties(dataset, concept)
+            try:
+                r['schema'] = self.update_properties(dataset, concept)
+            except requests.exceptions.HTTPError as e:
+                if e.response.status_code == 400:
```
mikeswoods: Are there any other 4XX codes that we need to handle? Otherwise looks good

mattusifer: I don't think so, I am trying to handle [these](https://github.com/Blackfynn/blackfynn-api/blob/DEVELOPMENT/concepts-core/src/main/scala/com/blackfynn/concepts/managers/ConceptPropertiesManager.scala#L85) errors from the concepts service, each of which signal that the properties were invalid. All of these errors are [mapped](https://github.com/Blackfynn/blackfynn-api/blob/DEVELOPMENT/concepts-core/src/main/scala/com/blackfynn/concepts/ConceptsApiError.scala#L24) to 400 Bad Request. 

Let me know if you think there are any other errors I should handle as part of this PR.

PR #39, opened 2018-07-27
APP-1041: Fix Annotation Layer Deletion

mattusifer:


mistyahmed:
Just to confirm, will the issue with deletion of annotations be addressed in a separate PR?

mistyahmed:
@mattusifer Thanks for the explanation!  I think you can make a similar change here: https://github.com/Blackfynn/blackfynn-python/blob/534b4f10270ba2f87288bbb34f7767c3f56e30cd/blackfynn/api/timeseries.py#L569

mattusifer:
Good catch! I'll make that change as part of this PR :+1:

mattusifer:
I noticed there were a number of commented-out tests that were supposed to be testing deletion functionality, so I simply uncommented those instead of writing new tests.

PR #40, opened 2018-07-27
APP-1019: Optionally include record id in as_dataframe

mattusifer:


blackfynn/models.py
```diff
@@ -3209,17 +3209,34 @@ def as_dataframe(self):
...
+
         """
         cols = self.type.schema.keys()
+
+        if record_id_column_name:
+            if record_id_column_name in cols:
+                raise ValueError("There is already a column called '{}' in this data set.".format(
+                    record_id_column_name
+                ))
+            cols = [record_id_column_name] + cols
```
mikeswoods: Totally optional: alternatively, you can update in place with `cols.insert(0, record_id_column_name)` since, `cols` is being reassigned anyway

mattusifer: I like that idea :+1:

PR #53, opened 2018-08-03
Upload without extra platform requests

rohanshah:
# Description
In the main upload function we re-fetch objects (dataset and destination package) in order to verify that they are actual objects that exist on the platform. While this is good validation, performing a large number of uploads to the same destination and dataset requires needlessly making requests from the platform. Further, when the destination is collection, the API always responds with its constituent children, which, when continuously uploading to, can cause the time of that request to increase to the point where it may timeout.

## Type of change

Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

Unit tests plus used this branch for uploading PPMI MRI data to the platform.

# Checklist:


blackfynn/api/transfers.py
```diff
@@ -217,13 +218,19 @@ def upload_files(self, destination, files, dataset=None, append=False, display_p
 
         # sanity check dataset
         try:
-            ds = self.session.datasets.get(dataset_id)
+            if isinstance(dataset, Dataset) and dataset.exists:
```
merky: If one has a destination other than a Dataset, this will fail, since `dataset` will not be set.

merky: Oh, nevermind! I read this incorrectly.

PR #62, opened 2018-08-23
APP-1141: Update Python Client to Support New Datatypes

mistyahmed:
# Description

The Blackfynn API now allows model properties to optionally be marked as `required`.  If a property is marked as required, all future model records must contain a value for that property.  The Python client has been updated to allow users to set the `required` field per property.

Additionally, the Blackfynn API now allows a `format` to be associated with string-type properties and a `unit` to be associated with numeric-type properties.  The Python client has been updated to allow users to specify a format or unit when creating a property definition.

## Type of change

# How Has This Been Tested?

New tests have been added to `tests/test_concepts.py` to validate that the `required` property field and complex property data types are handled as expected.

# Checklist:


blackfynn/models.py
```diff
@@ -2190,24 +2199,62 @@ def uncast_value(value):
...
+                unit=self.unit
+            )
+        else:
+            d = convert_type_to_datatype(self.data_type)
+
+        return d
 
 class BaseModelProperty(object):
-    def __init__(self, name, display_name=None, data_type=basestring, id=None, locked=False, default=True, title=False, description=""):
+    def __init__(self, name, display_name=None, data_type=BaseModelPropertyType(data_type=basestring), id=None, locked=False, default=True, title=False, description="", required=False):
```
merky: You shouldn't be calling functions/creating objects in your `__init__` definition. I would just leave this as `data_type=str` for clarity and convention (it'll get wrapped later).

blackfynn/models.py
```diff
@@ -2139,24 +2139,33 @@ def as_dict(self):
     'date': datetime.datetime
 }
 
-def parse_model_datatype(data_type=None):
+def convert_type_to_datatype(model_type):
+    if model_type is None or isinstance(model_type, type) and model_type in model_type_map:
+        t = model_type_map[model_type]
```
mattusifer: If `model_type` is None, this will raise a `KeyError`

blackfynn/models.py
```diff
@@ -2190,24 +2199,62 @@ def uncast_value(value):
...
+        self.format = format
+        self.unit = unit
+
+    @classmethod
+    def from_dict(cls, data):
+        data_type = data.get('data_type', data.get('dataType'))
+
+        if isinstance(data_type, dict):
+            t = data_type['type'].lower()
+            f = data_type.get('format', None)
```
mattusifer: No need to default to `None`

blackfynn/models.py
```diff
@@ -2190,24 +2199,62 @@ def uncast_value(value):
...
 class BaseModelProperty(object):
-    def __init__(self, name, display_name=None, data_type=basestring, id=None, locked=False, default=True, title=False, description=""):
+    def __init__(self, name, display_name=None, data_type=BaseModelPropertyType(data_type=basestring), id=None, locked=False, default=True, title=False, description="", required=False):
         assert ' ' not in name, "name cannot contain spaces, alternative names include {} and {}".format(name.replace(" ", "_"), name.replace(" ", "-"))
 
         self.id = id
         self.name = name
         self.display_name = display_name or name
-        self.type = parse_model_datatype(data_type)
+        self.data_type = wrap_datatype(data_type)
```
merky: I understand that you want backwards compatibility, so you left `self.type` (simple) and created a new `self.data_type` (wrapped). However, you should just maintain a central source of `data_type` without duplicating it. For example:

```python
def __init__( ... ):
    # ...
     self._type = wrap_datatype(data_type)
    # ...
@property
def type(self):
    return convert_datatype_to_type(self._type)
```
That way, you maintain backwards compatibility with `.type`, you can use `._type` to get full object, but you have a single source for the data type (not duplicated). 

Alternatively! You can just go ahead and make `.type` point to the full object. To help with backwards compatibility, you could then override the `__eq__` method for the type class, so that it would be equivalent to, e.g. `(str, basestring, "string")`, etc. The API changes are going to break the client regardless, and folks are gonna have to update to interact with the concepts service anyway 🤷‍♂️ .

blackfynn/models.py
```diff
@@ -2190,24 +2199,62 @@ def uncast_value(value):
 
     return v
 
+class BaseModelPropertyType(object):
```
merky: It's not 100% clear to me why this is a `Base` class vs. just `ModelPropertyType`. Is it used more than once?

blackfynn/models.py
```diff
@@ -2190,24 +2199,62 @@ def uncast_value(value):
...
+    def __init__(self, data_type, format=None, unit=None):
+        self.data_type = data_type
+        self.format = format
+        self.unit = unit
+
+    @classmethod
+    def from_dict(cls, data):
+        data_type = data.get('data_type', data.get('dataType'))
```
merky: I would remove this extraction process, and place it back in the `ModelValue.from_dict()`, etc. It's ideal that the `data` argument entering this function represents the data specifically for `ModelPropertyType`, not a more encompassing object.

blackfynn/models.py
```diff
@@ -2308,7 +2362,7 @@ def __init__(self, dataset_id, name, display_name = None, description = None, lo
 
         self._add_properties(schema)
 
-    def _add_property(self, name, display_name=None, data_type=basestring, title=False):
+    def _add_property(self, name, display_name=None, data_type=BaseModelPropertyType(data_type=basestring), title=False):
```
merky: Same here; avoid calling stuff inside `__init__`.

blackfynn/models.py
```diff
@@ -2343,7 +2397,7 @@ def _validate_values_against_schema(self, values):
     def update(self):
         pass
 
-    def add_property(self, name, data_type=basestring, display_name=None, title=False):
+    def add_property(self, name, data_type=BaseModelPropertyType(data_type=basestring), display_name=None, title=False):
```
merky: And here.

blackfynn/models.py
```diff
@@ -2554,6 +2608,10 @@ class ModelProperty(BaseModelProperty):
     def __repr__(self):
         return u"<ModelProperty name='{}' {}>".format(self.name, self.type)
 
+class ModelPropertyType(BaseModelPropertyType):
```
merky: I said this above, but I think you could just move everything into `ModelPropertyType` and be done with it. No base class necessary. Let me know if I'm missing something!

tests/test_concepts.py
```diff
@@ -150,3 +190,86 @@ def test_models(dataset):
...
+
+     updated_model_1 = dataset.get_model(model_with_basic_props_1.id)
+
+     assert updated_model_1.get_property('a_new_property').display_name == 'Weight'
+
+     # Define properties as ModelProperty objects
+     model_with_basic_props_2 = dataset.create_model('Basic_Props_2', description='a new description', schema=[
+         ModelProperty('name', data_type=str, title=True, required=True),
+         ModelProperty('age', data_type=int),
+         ModelProperty('DOB', data_type=long)
```
merky: I mentioned this above, but we should also test for `data_type='date'`, `data_type='string'`, etc.

blackfynn/models.py
mattusifer: where do these two formats come from?

blackfynn/models.py
```diff
@@ -2219,33 +2266,40 @@ def from_tuple(cls, data):
...
-        return cls(name=data['name'], display_name=display_name, data_type=data_type, id=id, locked=locked, default=default, title=title)
+        return cls(name=data['name'], display_name=display_name, data_type=data_type, id=id, locked=locked, default=default, title=title, required=required)
 
     def as_dict(self):
         return dict(
             id           = self.id,
             name         = self.name,
             displayName  = self.display_name,
-            dataType     = convert_datatype_to_model_type(self.type),
+            dataType     = self.data_type.as_dict(),
```
mattusifer: Any idea why these aren't using snake case? I know this precedent was already set before this PR, just looks odd.

tests/test_concepts.py
```diff
@@ -150,3 +190,86 @@ def test_models(dataset):
...
+    assert (propertyType.data_type == unicode)
+    assert (propertyType.format.lower() == 'email')
+    assert (propertyType.unit == None)
+
+    good_values = {'name': 'Bob', 'age': 1, 'email': 'test@test.com', 'weight': 10}
+    good_record = updated_model.create_record(good_values)
+
+    bad_values = {'name': 'Bob', 'age': 1, 'email': '123455', 'weight': 10}
+
+    try:
```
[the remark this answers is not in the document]
mistyahmed: Thanks, I've made the suggested change.

blackfynn/models.py
```diff
@@ -2139,24 +2139,33 @@ def as_dict(self):
     'date': datetime.datetime
 }
 
-def parse_model_datatype(data_type=None):
+def convert_type_to_datatype(model_type):
```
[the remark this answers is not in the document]
mistyahmed: I've added a condition to `convert_type_to_datatype` that handles reverse look up of the property data type.

blackfynn/models.py
```diff
@@ -2139,24 +2139,33 @@ def as_dict(self):
...
+    if model_type is None or isinstance(model_type, type) and model_type in model_type_map:
```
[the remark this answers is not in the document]
mistyahmed: I've removed the `if model_type is None` condition from `convert_type_to_datatype` (as well as `if data_type is None` from `convert_datatype_to_type`) since I expect the type to always be defined.  Thoughts?

blackfynn/models.py
mistyahmed: This has been addressed.

blackfynn/models.py
mistyahmed: I've implemented your first suggestion.

blackfynn/models.py
mistyahmed: Nope, I've renamed `BaseModelPropertyType` to `ModelPropertyType`.

blackfynn/models.py
mistyahmed: This has been addressed.

blackfynn/models.py
mistyahmed: This has been addressed.

blackfynn/models.py
mistyahmed: This has been addressed.

tests/test_concepts.py
mistyahmed: I've added unit tests for these cases.

blackfynn/models.py
```diff
@@ -2139,24 +2139,33 @@ def as_dict(self):
...
-    if data_type is None:
-        return
-
-    assert isinstance(data_type, type) and data_type in model_type_map, "data_type must be one of {}".format(model_type_map.keys())
-
-    t = model_type_map[data_type]
-    return t.title()
+def wrap_datatype(data_type):
+    if isinstance(data_type, BaseModelPropertyType):
+        t = data_type
```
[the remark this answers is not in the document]
mistyahmed: Hmm I don't really understand.  Can you elaborate?

blackfynn/models.py
mistyahmed: Good point.  I've removed the conditions that check for `model_type/data_type is None` since I expect the type to always be defined.

blackfynn/models.py
mistyahmed: This has been addressed.

blackfynn/models.py
```diff
@@ -2301,6 +2304,10 @@ def as_dict(self):
     def as_tuple(self):
         return (self.name, self.type, self.display_name, self.title, self.required)
 
+    @property
+    def type(self):
```
merky: Since you're including the dynamic property `prop.type`, you may want to consider providing a corresponding "setter" method, so one could do this successfully: `prop.type = str` (or equivalent `prop.set_type(str)`) and it would properly update the `prop._type` object? Not needed necessarily, but is helpful whenever you have a dynamic property like this.

merky: Also, you may want to add a similar `.unit` method which would return the value in `._type.unit`, as I can see determining the unit of your property will be a common task.

blackfynn/models.py
mistyahmed: I believe it's because the json parser used by API expects the field names to be in camel case.  I tried changing these to camel case and confirmed that the API rejected the requests with a 'The request content was malformed' error.

blackfynn/models.py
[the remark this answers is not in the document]
merky: This is super minor - I was suggesting something like this:
```python
if isinstance(data_type, ModelPropertyType):
    return data_type
return ModelPropertyType(data_type=data_type)
```

blackfynn/models.py
mistyahmed: I've implemented both of these suggestions.

merky: 🎉

blackfynn/models.py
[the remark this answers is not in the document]
mistyahmed: Ah, ok.  I have implemented that change.

PR #66, opened 2018-08-30
APP-1184: Support model templates

mistyahmed:
# Description

The Python client now allows users to validate, create, retrieve, import, and delete model templates.

The following methods have been added to the `Blackfynn` class:
`get_model_template`
`get_model_templates`
`create_model_template`
`validate_model_template`
`delete_model_template`

The following method has been added to the `Dataset` class:
`import_model`

## Type of change

# How Has This Been Tested?

New integrated tests are included in `tests/test_templates.py`.

# Checklist:


tests/test_templates.py
```diff
@@ -0,0 +1,55 @@
...
+                type="number",
+                unit="lbs",
+                description="Weight"
+            )
+        ),
+        required=["name"]
+    )
+
+    # Validate template schema
+    validation_response = client.validate_model_template(template)
```
mattusifer: can we just add an assertion to set some expectations around this response?

tests/test_templates.py
```diff
@@ -0,0 +1,55 @@
+import pytest
+from blackfynn.models import ModelProperty, ModelTemplate
+
+def test_templates(client, dataset):
```
mattusifer: it would be useful to have an invalid template and some tests to establish expected error states.

blackfynn/client.py
```diff
@@ -295,5 +296,103 @@ def _check_context(self):
...
+        """
+        Create a new model template belonging to the current organization.
+
+        Args:
+            template (ModelTemplate): A ModelTemplate object.
+
+        Example::
+
+            my_template = ModelTemplate(
+                schema="http://schema.blackfynn.io/model/draft-01/schema",
```
merky: There are a lot of required fields here. Can we help the user out by having sane default, and/or additional logic to help make a template simpler? 

Examples:
- `schema` will almost always be `"http://schema.blackfynn.io/model/draft-01/schema"` and thus shouldn't be needed by the user. 
- `description` could be optional, and just set to the `name` if left blank
- It would be nice if `properties` could accept varying input structures. I would look at how you can specify properties on `Model` for inspiration, e.g. `properties=[('Name', str), ('Date', 'date')]` or `properties=dict(name=str, date='date')`.

merky: Also, for this specific method `create_model_template`, I also like the idea of having the arguments serve as a direct interface to the model, rather than having the argument be a single model. 

For example, `bf.create_model_template('Patient', category='Person', properties=[('Name', str), ('DOB', 'date')])` would be nice.

Or, for creating templates directly from a model, you could do:
```python
model =  ModelTemplate(...)
bf.create(model)
```

blackfynn/models.py
```diff
@@ -2047,6 +2047,19 @@ def create_relationship_type(self, name, description, schema=None, **kwargs):
                 **kwargs)
         return self._api.concepts.relationships.create(self.id, r)
 
+    def import_model(self, template):
```
merky: Is it possible that input could be `model_template_or_model` ? It'd be _awesome_ to be able to import a template directly from an existing `Model` type (from another dataset). That, or have another function `import_template_from_model(...)`? Just some thoughts 😄

mistyahmed: Good idea.  I think I'm going to tackle that in a separate PR.

blackfynn/client.py
mistyahmed: I believe this feedback has been addressed.

blackfynn/client.py
```diff
@@ -318,12 +318,25 @@ def get_model_templates(self):
...
+                              required=None,
+                              properties=None,
+                              template=None
+                              ):
         """
         Create a new model template belonging to the current organization.
 
         Args:
             template (ModelTemplate): A ModelTemplate object.
+            OR
```
merky: I would put a test assertion here (or in the called function) to make sure the user is adhering to one of the two ways: 1) template, or 2) name, description, etc.

blackfynn/api/concepts.py
```diff
@@ -438,3 +438,52 @@ def create(self, dataset, external_id, relationship, concept_instance, values, d
...
+        org_id = self._get_int_id(self.session._context)
+        resp = self._get(self._uri('/organizations/{orgId}/templates/{templateId}',
+                                   orgId=org_id, templateId=template_id))
+        return ModelTemplate.from_dict(resp)
+
+    def validate(self, template):
+        assert isinstance(template, ModelTemplate), "template must be type Template"
+        return self._post(self._uri('/validate'), json=template.as_dict())
+
+    def create(self, name, description, category, required, properties, template=None):
```
merky: You've mimicked the dynamic argument structure as `import_model_template` method on the dataset, which is fine, but since the docstrings are not on this method (they are on the model) it becomes a bit confusing what's needed/required to make the API call. Personally, I would either

1. Revert the input back to just `template` and do that logic in the model method
2. Create two functions here `create_from_template(template)` and `create_from_params(name, description, ...)` which are called conditionally elsewhere, or
3. Simply add docstrings here that clarify the input (essentially what is on the model)

blackfynn/client.py
```diff
@@ -295,5 +296,142 @@ def _check_context(self):
...
+            Object of type ModelTemplate.
+        """
+        return self._api.templates.get(template_id)
+
+    def get_model_templates(self):
+        """
+        Get all of the model templates belonging to the current organization.
+
+        Returns:
+            Objects of type ModelTemplate.
```
merky: Does this return a `list`?

blackfynn/models.py
merky: Makes sense!

blackfynn/client.py
```diff
@@ -295,5 +296,142 @@ def _check_context(self):
...
+        return self._api.templates.create(
+            name=name,
+            description=description,
+            category=category,
+            required=required,
+            properties=properties,
+            template=template
+        )
+
+    def validate_model_template(self, template):
```
merky: I just realized that this is not a dataset-specific action. Should this be moved to the `Blackfynn` object directly?

merky: Additionally, we should maybe return `True` if successful so that this could be used programmatically. Checking for `"Success"` doesn't seem pythonic.

blackfynn/models.py
```diff
@@ -2617,6 +2630,53 @@ def as_dict(self):
     def __repr__(self):
         return u"<BaseRecord type='{}' id='{}'>".format(self.type, self.id)
 
+
+class ModelTemplate(BaseNode):
+    _object_key = None
+
+    def __init__(self, name, category, properties, id=None, display_name=None, schema=None, description=None, required=None,
```
merky: Should we test for "required" inputs here? Thus, being only able to create a local `ModelTemplate` object if they satisfy the right fields. Helping the user out locally as much as possible (rather than relying on some API error) is always the preferred route.

blackfynn/api/concepts.py
mistyahmed: Thanks, I implemented your first suggestion.

blackfynn/client.py
mistyahmed: `validate_model_template` is on the Blackfynn object.  Also, I've updated that method to return `True` upon successful validation.

blackfynn/client.py
mistyahmed: Yes, I've updated the documentation to reflect that.

tests/test_templates.py
```diff
@@ -0,0 +1,88 @@
...
+                description="Weight"
+            )
+        ),
+        required=["name"]
+    )
+
+    # Validate template schema
+    validation_response_1 = client.validate_model_template(template=template_1)
+
+    assert(validation_response_1 == True)
```
mattusifer: This will be ```{message: 'Validation was successful!'}```

blackfynn/api/concepts.py
```diff
@@ -462,7 +462,7 @@ def get(self, template_id):
     def validate(self, template):
         assert isinstance(template, ModelTemplate), "template must be type Template"
         response = self._post(self._uri('/validate'), json=template.as_dict())
-        if "Validation was successful" in response:
+        if "Validation was successful" in str(response):
```
merky: @mistyahmed Does this need to be `return "Validation was successful" == response.get('message')` ?

PR #67, opened 2018-09-05
APP-1183: Add support for array and enum-type Model properties

mistyahmed:
# Description

The Blackfynn API now allows users to associate an enumeration of acceptable values with a ModelProperty.  Properties of type `array` allow multiple values from the property enumeration to be attached to a record.  Properties of type `enum` allow a single value from the property enumeration to be attached to a record.

The Python client has been updated to allow users to get and create model properties of types `array` and `enum`.

## Type of change

# How Has This Been Tested?

New tests have been added to `tests/test_concepts.py` to validate that model properties of type `enum` and `array` are parsed and handled appropriately.

# Checklist:


blackfynn/models.py
```diff
@@ -2205,43 +2205,74 @@ def uncast_value(value):
     return v
 
 class ModelPropertyType(object):
-    def __init__(self, data_type, format=None, unit=None):
+    def __init__(self, data_type, format=None, unit=None, enum=None, collection_type=None):
```
merky: Seems like we're mixing type complexities here. I think I'd advocate for making a new derived/extended class, e.g. `ModelPropertyEnumType`, which has the traditional `ModelPropertyType` as a base. This class could then have all the properties/methods specific to a collection. This removes the need for all the conditional logic below.

blackfynn/models.py
```diff
@@ -2205,43 +2205,74 @@ def uncast_value(value):
...
+        assert(collection_type is None or collection_type == 'array' or collection_type == 'enum',
+               "Collection type {} is not supported. Try 'enum' or 'array'".format(collection_type))
         self.data_type = data_type
         self.format = format
         self.unit = unit
+        self.enum = enum
+        self.collection_type = collection_type
+
+    def is_complex(self):
```
merky: `complex` seems like a confusing term here — is this primarily use for your conditional logic within the class, or is it intended for something else?

blackfynn/models.py
```diff
@@ -2205,43 +2205,74 @@ def uncast_value(value):
...
+                enum = None
+                collection_type = None
         else:
             t = data
-            f = None
-            u = None
+            format = None
+            unit = None
+            enum = None
+            collection_type = None
```
merky: I'm not sure what `collection_type` is. Or, rather, what is `array` vs. `enum`? From the looks of it (your test cases), these are all `enum` types, where there is some finite collection of options to choose from, no? I may be totally confused, please clarify 😄

blackfynn/models.py
mistyahmed: `is_complex` is used to determine how to serialize the property so that the concepts service can parse it.  The concepts service expects "simple" property types to look like this:
```
"dataType": "Double"
```
But "complex" property types should look like this:
```
"dataType": {
  "type": "Double",
  "unit": "cm"
}
```
"Complex" property types have additional fields (like `format` and `unit`) nested inside the `dataType` field.  Perhaps I can change the method name to `is_nested`?

blackfynn/models.py
mistyahmed: If the `collection_type` is `array`, then the property allows multiple values from the property enumeration to be attached to a record. If the `collection_type` is `enum`, then the property allows a single value from the property enumeration to be attached to a record.
`array` and `enum` are the only possible collection types.

blackfynn/models.py
merky: This seems like an internal-only method then, yes? And, since this is specific to the `to_dict` method, I would just include it in there, or... rename to `_is_complex_format` or whatever you wanna call it.

blackfynn/models.py
merky: I see. But, `array` can only be on type `enum`? What if I wanted to make a generic `int` type that is also an `array`? That is, without specifying the entire range of integers in some `enum`?

merky: Sorry, still a little confused by this: Wouldn't `collection_type` only ever be an `array`? When you create a record that has `enum` property type, you can only choose a single value for that property. It seems like there is some super-type that would enable the property to be a list of whatever the base type is (e.g. a `List` type with a base property that is `str` and `enum`).

mistyahmed: Regarding your first question, we don't plan to allow users to freely enter multiple values for a field, they can only select multiple values from a predefined list (if the property definition allows for that).

Maybe `collection_type` isn't the best terminology.  Would `selection_type` be better?  The `enum` field on the property type only contains the acceptable values for that field.  The `selection_type` field indicates whether the field allows multi-select.

mistyahmed: Looking again, I could replace `collection_type` with a boolean `multi_select` field and then just map that to `array` or `enum` appropriately. 🤔

blackfynn/models.py
mistyahmed: Thanks, I've implemented your suggestion.

blackfynn/models.py
```diff
@@ -2215,33 +2215,74 @@ def from_dict(cls, data):
...
-        return u"<ModelPropertyType data_type='{}' format='{}' unit='{}'".format(self.data_type, self.format, self.unit)
+        return u"<ModelPropertyType data_type='{}' format='{}' unit='{}'".format(self.data_type, self.format,
+                                                                                           self.unit)
+
+class ModelPropertyEnumType(ModelPropertyType):
+    def __init__(self, data_type, format=None, unit=None, enum=None, multi_select=False):
+        super(ModelPropertyEnumType, self).__init__(data_type, format, unit)
+        self.enum = enum
+        self.multi_select = multi_select
+        if self.multi_select is True:
```
merky: Just FYI, could write as `self.selection_type = 'array' if self.multi_select else 'enum'`

blackfynn/models.py
```diff
@@ -2215,33 +2215,74 @@ def from_dict(cls, data):
...
+            self.selection_type = 'array'
+        else:
+            self.selection_type = 'enum'
+
+    @classmethod
+    def from_dict(cls, data):
+
+        selection_type = data['type'].lower()
+        if selection_type == 'array':
```
merky: Same here, `multi_select = selection_type == 'array'`

PR #70, opened 2018-09-18
AE-130: Raise exception when creating dataset with existing name

rlmv:
# Description

Changes the behavior of `Client.create_dataset(name)` to raise an exception when a dataset with the given name already exists.

## Type of change

# How Has This Been Tested?

Added a test to `test_basic.py`. All tests pass locally.

# Checklist:


blackfynn/api/data.py
```diff
@@ -69,14 +69,11 @@ def create(self, ds):
         """
         Create a dataset on the platform
         """
-        try:
-            if self.get_by_name_or_id(ds.name) is not None:
-                raise Exception("Dataset with name {} already exists".format(ds.name))
-
-            resp = self._post('', json=ds.as_dict())
-            return Dataset.from_dict(resp, api=self.session)
-        except Exception, e:
```
mattusifer: I _think_ this was probably intended to catch exceptions from `self._post` in order to log errors that came out of the api. If we want to maintain that behavior we should probably leave this `except` here, make the exception type more specific than `Exception`, and raise a `ValueError` for duplicate names instead of an `Exception` above to make that more relevant as well.

If we're just going to `raise` on errors that come from `self._post` I'm fine with that too, as long as it's deliberate! Interested in @merky's take on this.

rlmv:
What's the preferred convention with logging and errors? Should this keep that log statement in there, or is it sufficient to just raise the exception?

blackfynn/api/data.py
rlmv: Gotcha. None of the other API calls in the module catch or log exceptions, so it's definitely an outlier in terms of behavior. In that case it might make sense to standardize that behavior across methods.

mattusifer: Sounds good, if other similar methods are doing it this way than I'm fine with it 👍

merky: @mattusifer In order to capture responses from the API, one would typically enable debugging via 
```shell
export BLACKFYNN_LOG_LEVEL=DEBUG
```
or equivalent. I _believe_ this would cover the use-case for the `try`/`catch` — if that's the case, the current PR is 👍

merky:
@mattusifer @rlmv There's definitely a larger conversation that could be had about surfacing better error messages in most cases, e.g. `ValueError` or customized error in this case. This would involve wrapping most model methods and/or API methods in the try/catch pattern that was used here, but also raising something more relevant. Perhaps we should add a ticket for this?

rlmv:
@merky Yes, I think it would be good to have more Blackfynn-specific error handling. I think `Exception` should generally not be raised directly, just for clarity's sake. We could add a `BlackfynnException` or something of the sort. I agree, probably a conversation for another thread.

blackfynn/api/data.py
merky: I think raising the error directly, without logging, is the way to go. So, you're good 👍

PR #71, opened 2018-09-18
Add support for topology endpoint

zduey:
# Description

The python client did not have a way of hitting the `/datasets/{datasetId}/concepts/{id}/topology` endpoint. This PR adds a method to the `Model` class that allows the user to return a list of all directly related models and the counts of those relationships. This corresponds to the `schemaRelationship`s in the topology-level graph. 

Added a utility module for the tests to hold any test-specific functions. Added a function to build a simple test graph that can be reused so we can keep the tests DRY especially as we add additional endpoints related to summarizing and querying the graph.

## Type of change

# How Has This Been Tested?
Ran unit tests locally

# Checklist:


tests/test_utils.py
```diff
@@ -0,0 +1,38 @@
...
+Graph = namedtuple('TestGraph', ['dataset', 'models', 'model_records',
+                                 'relationships', 'relationship_records'])
+
+
+def current_ts():
+    """ Gets current timestamp """
+    return int(round(time.time() * 1000))
+
+
+def build_simple_graph(dataset):
```
mattusifer: does this actually create any data in blackfynn? If so we might want to use setup/teardown to cleanup after tests that use this.

also, would it make sense to add this graph as a new fixture in `conftest.py`? That way it doesn't need to be recreated each time.

zduey: Yes, it does create data in Blackfynn. However, it does it using the `dataset` fixture, which gets cleaned up already (I think?). And yes, making it a fixture makes a lot of sense, although I don't think it should be done in `conftest.py` since it is specific to the tests within `test_concepts.py`. Does it make sense to you to have it as a module-level fixture inside of `test_concepts.py`?

blackfynn/api/concepts.py
```diff
@@ -147,6 +147,19 @@ def files(self, dataset, concept, instance):
             ))
         return [DataPackage.from_dict(pkg, api=self.session) for r,pkg in resp]
 
+    def get_related(self, dataset, concept):
+        assert isinstance(concept, Model), "concept must be type Model"
```
merky: Technically, you don't need this assertion, as `concept` could just be an ID (since line 153 exists).

blackfynn/models.py
```diff
@@ -2931,6 +2931,18 @@ def delete_records(self, *records):
         for error in result['errors']:
             self._logger.error("Failed to delete instance {} with error: {}".format(error[0], error[1]))
 
+    def get_related(self):
```
merky: Should we change this to `get_connected`? I think we talked about that this terminology being better. Then again... there are existing `get_related` methods elsewhere, [see here](https://github.com/Blackfynn/blackfynn-python/blob/master/blackfynn/models.py#L2960).

blackfynn/models.py
```diff
@@ -2931,6 +2931,18 @@ def delete_records(self, *records):
...
+        """ Returns a list of related model types and counts of those relationships
+
+        Returns:
+            List of ``Model``s
+
+        Example::
+            mouse.related()
```
merky: Mismatch in method name.

blackfynn/models.py
```diff
@@ -2931,6 +2931,18 @@ def delete_records(self, *records):
         for error in result['errors']:
             self._logger.error("Failed to delete instance {} with error: {}".format(error[0], error[1]))
 
+    def get_related(self):
+        """ Returns a list of related model types and counts of those relationships
```
merky: Define "related" here, perhaps?

tests/test_utils.py
```diff
@@ -0,0 +1,38 @@
+""" Utility functions for generating test fixtures """
```
merky: I think a good pattern is to only name files that have actual tests `test_<something>`.

tests/test_utils.py
merky: If the `graph` is going to be used in multiple tests, then yes, I would make it a fixture. 

Otherwise, it may be useful not to pull the graph build step out of that test file, because it's helpful to see what you're testing against. That is, it would be good to see in one spot (i.e. `test_get_related_models`) that you created two models, two records, and related those records, _before_ your actual test assertion.

blackfynn/models.py
zduey: I had the same thoughts go through my head and decided to be consistent with the existing naming conventions. I think we should rename these, but it should be done all at once probably in it's own PR

zduey:
Deleting this PR since it is part of #73. All suggestions from this thread have been incorporated into the other PR

PR #75, opened 2018-09-21
Show coverage results

rlmv:
# Description

Adds flags so pytest shows coverage statistics after each run.

## Type of change

# Checklist:


zduey:
Might not be a bad idea to look into using CodeCov on this repo so we can get automatic coverage statistics with each PR. Since it is public, I think we can use it for free.

rlmv:
I think you're right. It would be great to have those stats more publicly available. I'll investigate.

setup.cfg
```diff
@@ -4,3 +4,10 @@ universal = 1
 [metadata]
 description-file = README.rst
 license_file = LICENSE
+
+[tool:pytest]
+addopts = --cov=blackfynn --cov-report=term --cov-report=html
+
+[coverage:run]
+source = blackfynn
+omit = blackfynn/cli/*
```
merky: Good catch 👍

PR #79, opened 2018-09-26
AE-308 Improve request/retry handling

rlmv:
# Description

Improve retry handling to help with 504 Gateway Timeouts. We're still waiting on upstream rate-limiting to make this fully effective.

This also fixes an issue where a new session was created for every request.

## Type of change

# How Has This Been Tested?

All unit tests pass.

# Checklist:


blackfynn/base.py
```diff
@@ -104,10 +106,21 @@ def session(self):
         """
         Make requests-futures work within threaded/distributed environment.
         """
-        if not hasattr(self._session, 'session'):
```
merky: I'm assuming the reason we checked for `session` variable within the `requests.Session` object was because of the old threaded setup -- and it no longer applies here?

blackfynn/base.py
```diff
@@ -104,10 +106,21 @@ def session(self):
...
+        if self._session is None:
             self._session = Session()
             self._set_auth(self._token)
 
+            # Enable retries via urllib
+            adapter = HTTPAdapter(
+                max_retries=Retry(
+                    total=self.settings.max_request_timeout_retries,
+                    backoff_factor=.5,
+                    status_forcelist=[504] # Retry 504 errors (but not POSTs)
```
merky: This approach to generic retries is much nicer 👍. What about the other status codes, though?

blackfynn/base.py
rlmv: Yeah, the old FuturesSession` had a `session` attribute, but since we're using the vanilla `Session` that doesn't apply anymore. I'm not sure if there are any other implications for threaded environments (the old commit messages mention Celery specifically) but I tried using a thread pool to send concurrent requests with no issue.

blackfynn/base.py
rlmv: Maybe 503 and 502 as well? I think we generally don't want to retry 500s, but those others might be a good place to start.

rlmv: I'll add those

blackfynn/base.py
merky: Excellent!

merky: For historical context: we used to use our client _within_ our ETL process (which used to run on top of Celery).

rlmv: Aha, interesting!

PR #81, opened 2018-09-26
AE-39: Support Python 3

rlmv:
# Description

This is work in progress for Python 3 support. 

All code changes are complete. There's a fair amount of noise because I used isort to inject the required futures imports into all files and it had the side effect of sorting and cleaning up all the imports. If that's a concern I can pull those changes out.

### Remaining tasks:

- [x] Check integer/float division compatibility
- [ ] Write up developer information for 2/3 support
- [ ] Use `isort` (or something else) to ensure imports from `builtins` and `__futures__` are used everywhere
- [x] Update docs

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

# How Has This Been Tested?

# Checklist:


rlmv:
Yep, definitely would love to have some more eyes on this! This morning I realized that the unicode/string support is not quite right, but hopefully I should have that fixed by the end of the day.

blackfynn/api/base.py
```diff
@@ -25,7 +38,7 @@ def _get_id(self, thing):
         """
         Get ID for object. Assumes string is already ID.
         """
-        if isinstance(thing, (basestring, int, long)):
+        if isinstance(thing, (basestring, int)):
```
zduey: I think for all of these `isinstance()` calls where one of the things to check is a string type,  you want to be using `six`":
```python
from six import string_types
isinstance(s, string_types)
```

tox.ini
```diff
@@ -0,0 +1,6 @@
+[tox]
+envlist = py27,py34,py35,py36
+
+[testenv]
+deps = -rrequirements-test.txt
```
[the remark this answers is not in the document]
rlmv: Nope, it looks weird but it's valid tox syntax. I have no idea why. Maybe it does work without the space too?

blackfynn/api/base.py
rlmv: `past.builtins.basestring` does the same thing - that's where this basestring is coming from. It is a little confusing keeping track of the namespace overrides - that might be a benefit to using six instead of python-future.

zduey: Ah, gotcha. 👍

rlmv: Actually you're right, they're not identical. We actually don't want the `past.builtins.basestring` behavior in Python 3, `string_types` is the way to go. Good catch!

Issue #60, opened 2018-08-20
Accept file list in upload methods


merky:
**Client version:**

Version `2.4.4`

**I'm submitting a ...**(check one with "x")

* [ ] bug report
* [x] feature request

**Current behavior:**

File uploads into a dataset/collection using the method `dataset.upload` or `collection.upload` take a list of files as separate arguments (e.g. `dataset.upload('file1.dcm', 'file2.dcm', 'file3.dcm')`), but do not actually accept a list object directly as an argument. Thus, with a list `my_files` you can currently upload using `dataset.upload( *my_files )` where the `*` operator expands the list into arguments, but you cannot currently run the expected syntax of `dataset.upload(myfiles)`. 

**Expected behavior:**

Datasets and collections accept a python list-like object as an argument in the `.upload(...)` method.

rlmv:
Resolved by f96ceaba95d667dd9a03962b52f17a86b731f3a6

PR #87, opened 2018-10-02
AE-322: Upload list of files

rlmv:
# Description

Allow upload and append methods to either accept many file arguments, eg:

    dataset.upload(file1, file2, ...)

or take a list of files as a single argument:

    dataset.upload([file1, file2, ...])

## Github Issue

#60 

## Type of change

# How Has This Been Tested?

Updated existing upload and append tests to cover this option.

# Checklist:


blackfynn/models.py
```diff
@@ -750,6 +772,7 @@ def set_view(self, *files):
         Set the object(s) used to view the package, if not the file(s) or source(s).
         """
         self._check_exists()
+        files = _flatten_file_args(files)
```
merky: I don't believe you need/want to change anything for `set_view`, `set_files`, `set_sources`, or `append_to_sources`. These operate on `File` objects (which are objects that point to S3 locations for assets), not local files.

Actually, I'm not sure one can do this at all anymore, as this library is no longer used in the ETL and the API has restrictions on how these are set. @rohanshah can you confirm? If so, that means we should probably add another ticket for removing the `set_*` and `append_to_sources` methods altogether.

tests/test-upload-2.txt
```diff
@@ -0,0 +1 @@
+i am the data
```
merky: We should probably place all "test data" in a separate directory perhaps, `tests/resources`?

rlmv: Agreed

blackfynn/models.py
rlmv: Aha, interesting. Ok, I'll remove that. I do see now that those methods are excluded from the public API

zduey:
@merky @rlmv I have some code sitting around in the graphingest/ repo that recursively uploads from a directory, creating the corresponding folders/files on Blackfynn. This seems to be an extension of this PR. Is it worth pulling those over into the client?

tests/test_io.py
```diff
@@ -5,21 +5,32 @@
 from blackfynn.models import DataPackage, TimeSeries
 
 
-def test_upload(client, dataset):
+def _resource_path(fname):
```
zduey: There is a slightly "better" (IMHP) way of accessing test data. There is an example in this repo: https://github.com/Blackfynn/graph-ingest/blob/master/graphingest/test/test_dataset.py

It uses the `pkg_resources` module and sets up your `resources` folder so that it looks like just another package in your project so you can access that data without needing to do `os.path` stuff

rlmv:
@zduey Interesting!  Maybe that should be another ticket/PR - I could imagine some interesting corner cases there that might need some attention, eg around symlinks. Are you picturing that as a separate endpoint, eg. `upload_directory`?, or as an extension of `upload`?

tests/test_io.py
merky: Agree with @zduey, the `pkg_resources` approach looks like the way to go.

merky:
@zduey I wouldn't say it's a direct extension. This PR is to simply accept parameters as `upload(thing1, thing2)` *and* `upload([thing1, thing2])`.

However, yes, uploading a directory (syncing local hierarchy) is useful functionality. Technically, the python does this currently, but it is limited to the CLI (which will be depreciated anyway), and not in the actual library. I don't see any harm in bringing that logic over, and simply making our `thing.upload(...)` method more robust to file & directories, and recursive replication. This should be a separate ticket/PR though. 

In addition, there was some initial work to enable an "upload a directory" functionality in the API/front-end this past summer. I'm not sure if it was fully implemented or not. @mistyahmed, can you enlighten us?

zduey:
@rlmv I think maybe a separate PR makes the most sense since you are right that there probably does need to be some care taken around symlinks. I'm not sure about the `upload_directory` versus `upload`. I tend to prefer not having functions be able to do too many different things (upload one file, many files, directories, etc). Maybe @merky has an opinion here?

blackfynn/models.py
```diff
@@ -65,6 +65,27 @@ def _update_self(self, updated):
...
+        dataset.upload([file1, file2])
+    """
+    if len(files) == 1 and not isinstance(files[0], string_types):
+        # single argument - is it iterable?
+        try:
+            files = list(files[0])
+        except Exception:
+            pass
+
+    return files
```
rohanshah: This seems generic enough to move to `utils` and make it agnostic to whether the list arguments are filenames (strings) or not.

Not a hard requirement to change though, do what's in your heart.

tests/test_io.py
zduey: Seems like this can easily go either way. You are right that it is primarily for accessing data files as part of a distribution. So, if we wanted to ship some sample data as part of the library in the future, `pkg_resources` would be the way to go, but it does seem like this isn't necessary for test resources. I looked at a few other popular libraries and they do not follow the pattern I suggested, so I'm fine with leaving thing as-is with access via `os.path` since there isn't any benefit in this case to switching.

blackfynn/models.py
rlmv: The string check is important because otherwise the cast to list will happily convert a string into a list of characters.

I think I'm going to keep it in models.py for now; we can always move it once it is used in another module.

tests/test_io.py
rlmv: I'm always down to try new things. Updated!

rlmv:
I added tickets to JIRA for removing the DataPackage methods and uploading recursively.

PR #96, opened 2018-10-18
Avoid failure on graph-view auto-refresh

merky:
When retrieving all views for a dataset (`ds.views()`) the client attempts to get the latest of each view. Upon getting the latest view, and no versions are available, it attempts to "refresh" the view to create a new instance. However, if the 
```
POST /analytics/organizations/<org>/datasets/<dataset>/views/<view>/instances
```
fails, this prevents one from retrieving the list of views. This is a temporary fix to avoid that failure.


rlmv:
Although, this will hide any errors when a user calls `latest()` on a single view instance - that could be an issue. Is the problem that the server is choking on to many instance requests?

merky:
@rlmv Added an argument `ignore_errors` that's used when we call `ds.views()` but otherwise will raise an error, e.g. `view.latest()`. And, not sure why the server doesn't like handling some of the view definitions, but you can recreate the issue:

```python
from blackfynn import Blackfynn
bf = Blackfynn('dev')
ds = bf.get_dataset('PPMI V02.1')
views = bf.views() # fails (without this PR)
```

PR #109, opened 2018-10-30
Include snapshot status.

merky:
Facts:
 - Graph Views API recently added `status` field to all `GraphViewSnapshot`s.
 - Graph Views API added a `/instances/latest` endpoint
 - Graph Views API accepts `status` parameter on both `/instances` and `/instances/latest` endpoints.

Changes:
 - Both `view.latest()` and `view.get_snapshots()` now accept an argument `status`
 - Values of `status` can be `"processing"`, `"failed"`, or `"success"`. In addition, if value is `"any"` or `None`, endpoint will disregard filter.
 - Added method `view.get_snapshot` that retrieves snapshot by `id`.


rlmv:
Looks good!

merky:
@rohanshah Awesome, I'll make that change now.

rohanshah:
@merky cool the service should now be using `ready`

PR #112, opened 2018-11-01
Added snapshot download method

merky:
- Added `snapshot.download()` method to retrieve data as file directly (without loading into pandas)
- New method accepts either a filename (string) or file-like object, which enables one to also download directly into memory if need-be.
- The method `snapshot.as_dataframe()` now uses the `self.download()` method internally.


blackfynn/models.py
```diff
@@ -3750,6 +3744,35 @@ def as_dataframe(self, columns=None):
...
+        """
+        url = self._api.analytics.get_presigned_url(self.dataset_id, self, format=format)
+
+        def download_file_contents(response, f):
+            for chunk in response.iter_content(chunk_size=16384):
+                if chunk:
+                    f.write(chunk)
+
+        with requests.get(url, stream=True) as r:
+            if isinstance(location, basestring):
```
rlmv: Use `string_types` instead of `basestring` for Python 3

tests/test_analytics.py
```diff
@@ -24,6 +27,14 @@ def graph_view(simple_graph):
...
     view.delete()
 
+@pytest.fixture(scope='module')
+def ready_snapshot(graph_view):
+    # wait for snapshot to process
+    start = time.time()
+    while graph_view.latest() is None and (time.time() - start < 30):
+        time.sleep(1)
+    return graph_view.latest()
+
```
rlmv: Nice!

PR #113, opened 2018-11-01
Use base node

rlmv:
# Description

Use `BaseNode` instead of `BaseRecord` for graph views; tidy kwarg usage when building requests and materializing objects.

## Jira Issue

https://blackfynn.atlassian.net/browse/DS-585?

## Type of change

# How Has This Been Tested?

# Checklist:


blackfynn/api/concepts.py
```diff
@@ -585,82 +575,78 @@ def create_view(self, dataset, name, root, include):
             else:
                 raise
```
zduey: I can't remember if there is a reason for this pattern. If not, should we just let `self._post` do the raising since we aren't saying anything more specific?

blackfynn/models.py
```diff
@@ -3609,15 +3609,15 @@ def as_dataframe(self):
...
 
-    def __init__(self, name, root_model, included_models, *args, **kwargs):
+    def __init__(self, name, root_model, included_models, dataset_id, created_at, *args, **kwargs):
         self.name = name
         self.root_model = root_model
         self.included_models = included_models
-
-        kwargs['type'] = 'GraphViewDefinition'
+        self.dataset_id = dataset_id
+        self.created_at = created_at
```
zduey: Is this something that has to be provided to the API? This seems like something that should be automatically generated. I don't remember clients needing to specify `created_at` for other types of data containers in the platform

blackfynn/models.py
```diff
@@ -3707,12 +3707,15 @@ def __repr__(self):
         return u"<GraphViewDefinition name='{}' id='{}'>".format(self.name, self.id)
 
 
-class GraphViewSnapshot(BaseRecord):
+class GraphViewSnapshot(BaseNode):
+    _object_key = None
```
zduey: What do these actually do again? I see empty strings, `None`, and non-empty string values for these member variables in different models and the inconsistency seems odd

blackfynn/models.py
rlmv: `created_at` exists on the `BaseDataNode` which underlies most of the Python models. But that also adds support for adding properties which isn't supported (yet) by views.

blackfynn/models.py
rlmv: It's the key in the response that contains the actual data to construct the model. The empty string and `None` do the same thing: take the whole response object as the data source.

blackfynn/models.py
rlmv: Maybe down the line views will be a first-citizen of the platform and that will be supported, but I think for now we shouldn't inherit from that base class

blackfynn/api/concepts.py
rlmv: It's being used to inject the body of the response into the Exception - not sure why that isn't automatically done by requests, but I agree the repetition is clunky. I can open another PR to address that.

blackfynn/models.py
zduey: Cool, didn't realize it was a requirement of inheriting from `BaseNode`. Makes sense.

PR #124, opened 2018-11-14
SHIELD-502: Return all time series annotations

rlmv:
# Description

Currently only 100 annotations are returned due to the built in limit/offset in the API. 

This repurposes the unused `get_annotations` API method to retrieve all annotations in batches.

## Jira Issue
https://blackfynn.atlassian.net/browse/SHIELD-502

## Type of change

# How Has This Been Tested?

Added a unit test.

# Checklist:


tests/test_timeseries.py
```diff
@@ -286,6 +286,35 @@ def _sort_counts(counts):
...
     assert not layer2.exists
 
+
+def test_timeseries_annotations_can_get_more_than_default_limit(timeseries):
+    ch = TimeSeriesChannel(
+        name='test_channel',
+        rate=256,
+        unit='uV',
+        start=0,
+        end=200)
```
merky: This is an interesting example, as the end time of your channel is technically before the completion of a single sample (since this is only 256Hz). Remember: start/end times are in microseconds, so perhaps it would be a more realistic test to 1) have the time span of the channel extend beyond a single sample, and 2) have example annotations extend across those samples. Not sure this technically provides any difference to the test, but is more realistic.

tests/test_timeseries.py
```diff
@@ -286,6 +286,35 @@ def _sort_counts(counts):
...
+    timeseries.add_channels(ch)
+
+    layer = TimeSeriesAnnotationLayer(
+        name="limit_layer", time_series_id=timeseries.id,
+        description="layer with many small annotations")
+    timeseries.add_layer(layer)
+
+    for i in range(200):
+        layer.add_annotations(TimeSeriesAnnotation(label='annotation_{}'.format(i),
+                                     channel_ids=ch.id, start=i, end=i+1))
```
merky: See my comment above, as start/end might want to be adjusted here.

tests/test_timeseries.py
rlmv: Ahh good catch, good catch. I'll update that.

PR #129, opened 2018-11-27
analytics api endpoint changes

zduey:
# Description

We did some small refactoring on the analytics service API side that changed the endpoints. These changes update the python client to be in step with those changes


zduey:
@merky Started a discussion on slack related to the endpoints. Going to go ahead and merge this for now since these endpoints are already live and I want to make sure we can still use the client to interact with the service.

PR #137, opened 2018-12-11
added get summary method to datasets

zduey:
# Description

Added a new endpoint to the concepts service  (`/concepts/graph/summary`) for getting some high-level summary information about a graph. 

## Type of change

# How Has This Been Tested?

Ran unit tests locally

# Checklist:


blackfynn/models.py
```diff
@@ -1936,6 +1936,10 @@ def get_topology(self):
         """
         return self._api.concepts.get_topology(self)
 
+    def get_summary(self):
```
rlmv: Is there a more descriptive name we could use? Like `graph_summary`?

rlmv: Yep, that works!. Still wondering about the `get` part because most of the other dataset methods named like that get a specific object, whereas methods like `models()` return all the things. I suppose statistics are somewhere in between so I'm happy either way, just curious about the previous naming choices.

PR #139, opened 2018-12-21
AP-19

zduey:
# Description

Adds workspaces to client and moves view/snapshot under workspaces instead of datasets. Small refactoring so `models.py` can start being split out into smaller modules.

## Github Issue

[AP-19](https://blackfynn.atlassian.net/browse/AP-19)

## Type of change

# How Has This Been Tested?
Ran all unit tests locally against DEV

# Checklist:


blackfynn/__init__.py
```diff
@@ -4,7 +4,7 @@
 from .client import Blackfynn
 
 # base models
-from .models import (
+from models.models import (
```
mikeswoods: Any reason to change the import path from `from blackfynn.models` to `blackfynn.models.models`?

blackfynn/utils.py
```diff
@@ -137,3 +137,11 @@ def generate_dataframe(minutes=2, freq=100):
         'sawtooth': generate_data(n, func='sawtooth'),
         'square': generate_data(n, func='square')
     }, index=ind)
+
+def sanitize_name(name):
```
mikeswoods: Would '#' or '@' be acceptable in a name? A whitelist/regex of allowable characters might be preferable here.

blackfynn/__init__.py
zduey: The `models.py` module was a few thousand lines long, so I turned it into a package and pulled out the workspaces and analytics stuff. The plan is to eventually pull out the remaining models into their logically-related groups and just have a`models_base` for the parent classes. I didn't want to do all of that refactoring in this PR, so this is just the barebones start that keeps most of the original `models.py` module.

blackfynn/utils.py
zduey: I just looked at I don't think this utility is even used any more. I'm going to confirm and then just rip it out.

blackfynn/__init__.py
rlmv: Could we move the old `models.py` into `models/__init__.py`? Then all the imports should be backwards-compatible.

rlmv: Or do an `from .models import *` in `models/__init__.py`?

blackfynn/models/workspace.py
```diff
@@ -0,0 +1,312 @@
...
+        )
+
+    @as_native_str()
+    def __repr__(self):
+        return u"<GraphViewDefinition name='{}' id='{}'>".format(self.name, self.id)
+
+
+class GraphViewSnapshot(BaseNode):
+    _object_key = None
+    _valid_snapshot_status = ['failed','processing','ready']
```
rlmv: I think this was like this before, but whitespace around list elements.

blackfynn/models/workspace.py
```diff
@@ -0,0 +1,312 @@
...
+    def __init__(self, name, description=None, **kwargs):
+        kwargs.pop('package_type', None)
+        kwargs.pop('type', None)
+        super(Workspace, self).__init__()
+        self.name = name
+        self.description = description or ''
+
+    def __eq__(self, other):
+        """ Workspaces within an organization are differentiated  by name """
+        return self.name == other.name
```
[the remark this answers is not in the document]
rlmv: Otherwise we _could_ compare two identically named workspaces in different orgs as identical, unlikely as that is.

blackfynn/__init__.py
zduey: Yeah, I'll give that a shot. I have a vague recollection of there being some issue with imports getting wonky with a package and a module that have the same name, but hopefully I'm wrong.

blackfynn/models/workspace.py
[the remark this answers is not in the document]
zduey: I'll switch it to be by id since name and id seems redundant

blackfynn/api/analytics.py
```diff
@@ -0,0 +1,132 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import absolute_import, division, print_function
+from builtins import object, zip
+from future.utils import as_native_str, string_types, PY2
+
+import pdb
```
rlmv: Extra import

blackfynn/api/workspaces.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+class WorkspacesAPI(APIBase):
+    """
+    Interface for managing workspaces on the platform.
+    """
+    base_uri='/workspaces'
+    name = 'workspaces'
+
+
```
rlmv: Extra whitespace

blackfynn/api/workspaces.py
```diff
@@ -0,0 +1,65 @@
...
+    def get(self, name_or_id):
+        workspace_id = name_or_id
+        if isinstance(name_or_id, (string_types)):
+            workspaces = self.get_all()
+            match = find_by_name(workspaces, name_or_id)
+            if match is None:
```
rlmv: We're hitting the API an extra time here - can this just be 
```
if match is not None:
    return match
```
Sorry I didn't catch this earlier.

blackfynn/models/__init__.py
```diff
@@ -0,0 +1 @@
+from .models import *
```
rlmv: I think this needs to be preceded by `from __future__ import absolute_import` for Python 2 compatibility

blackfynn/models/workspace.py
```diff
@@ -0,0 +1,311 @@
...
+        self.description = description or ''
+
+    def __eq__(self, other):
+        return self.id == other.id
+
+    @as_native_str()
+    def __repr__(self):
+        return u"<Workspace name='{}' id='{}'>".format(self.name, self.id)
+
+    def create_view(self, dataset, name, root, include, create_snapshot=True):
```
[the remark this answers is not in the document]
zduey: Yes, it ultimately gets passed to `_get_int_id` from `APIBase`, which handles the case of the thing being an object.

blackfynn/api/analytics.py
```diff
@@ -0,0 +1,131 @@
...
+                        **self._with_kwargs(workspace, dataset=dataset))
+
+        try:
+            resp = self._post(uri, json={
+                'name': name,
+                'rootModel': root,
+                'includedModels': include,
+            })
+        except requests.exceptions.HTTPError as e:
+            if e.response.status_code == 400:
```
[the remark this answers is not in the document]
zduey: @rlmv -- Any recollection why this was in there? Looks like this ultimately passes through `_handle_response` in the `BlackfynnRequest` class which calls `raise_for_status` from the requests library if there was an http error. Perhaps it wasn't raising the error with the informative text from the response and that's why this was there?

blackfynn/models/workspace.py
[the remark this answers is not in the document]
merky: 👍

blackfynn/models/workspace.py
```diff
@@ -0,0 +1,311 @@
...
+            name = self.name,
+            description = self.description,
+        )
+
+
+class GraphViewDefinition(BaseNode):
+    _object_key = None
+
+    def __init__(self, name, root_model, included_models, workspace_id,
+                 dataset_id, created_at, *args, **kwargs):
```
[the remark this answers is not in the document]
zduey: pop-ing `created_at` makes sense to me since that's the common pattern. As for the `root_model` and `included_models`, it takes the string names of the models for right now. I can see taking the model objects and pulling off the name from those. I'm not sure how much benefit accepting ids gives since the user is much more likely to know the name or have an object on hand than they are to have an id. At first thought, that seems like a bit too much flexibility. Happy to be convinced otherwise if there is a good use case.

blackfynn/models/workspace.py
```diff
@@ -0,0 +1,311 @@
...
+        try:
+            # TODO: allow for wildcard columns, e.g. `patient.*`
+            df = pd.read_parquet(filename, columns=columns)
+        except pyarrow.ArrowIOError:
+            with io.open(filename) as f:
+                self._check_response(f.read())
+            raise
+        finally:
+            try:
+                os.remove(filename)
```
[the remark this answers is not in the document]
zduey: In general, my opinion is that it shouldn't be the python client's job to maintain a cache. My rationale is that clients are usually just a thin wrapper around an API specific to a particular language and the more stuff we add to a specific client, the harder it is to keep all clients in-sync with each other. If the user wants to keep a hold of the data on disk, they have the dataframe and can always do that.

blackfynn/api/analytics.py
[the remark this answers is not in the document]
zduey: Separate PR makes sense. This one is already getting a little sprawling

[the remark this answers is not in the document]
zduey: This is now a ticket: https://blackfynn.atlassian.net/browse/SHIELD-587

PR #141, opened 2019-02-01
add status to datasets

natevecc:
# Description

Add status to dataset class

## Github Issue
N/A

## Type of change

# How Has This Been Tested?

Successfully interacted with new version of the API in dev environment 

# Checklist:


blackfynn/models.py
```diff
@@ -1885,11 +1885,12 @@ def __repr__(self):
 # ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 class Dataset(BaseCollection):
-    def __init__(self, name, description=None, **kwargs):
+    def __init__(self, name, description=None, status=None, **kwargs):
```
rlmv: I think you need to also add `status` to the `as_dict` method.

natevecc: Dataset doesn't have a `as_dict` method. Should I add it?

rlmv: It's just hiding right here https://github.com/Blackfynn/blackfynn-python/blob/fbf2366f50f1b729859f118c0eb6184d8f412030/blackfynn/models.py#L2099

rlmv: Gotcha, you should be good then

PR #144, opened 2019-03-01
Package Processing Functionality

rohanshah:
# Description

JIRA: [AP-526](https://blackfynn.atlassian.net/browse/AP-526)

Datasets now have a automatically process packages flag that indicates whether or not the dataset is set to automatically process new packages or keep them in a completed `UPLOADED` state. This adds the `automatically_process_packages` flag to the `Dataset` object to provide this information to the user. 

Packages that are not automatically processed can be processed by using the new process endpoint. This adds the `process` function to `DataPackage` objects to enable this functionality.

## Type of change

# How Has This Been Tested?

Testing via `ipython` to make sure that the endpoint is being called appropriately and reporting errors correctly.

# Checklist:


blackfynn/api/data.py
```diff
@@ -226,6 +227,23 @@ def get(self, pkg, include=None):
         pkg = self._get_package_from_data(resp)
         return pkg
 
+    def process(self, pkg):
+        """
+        Process a package that has been successfully uploaded but not yet processed
+        """
+        try:
+            self._put(self._uri('/{id}/process',id=pkg.id))
```
rlmv: whitespace before `id`, please.

blackfynn/api/data.py
```diff
@@ -226,6 +227,23 @@ def get(self, pkg, include=None):
...
+        except requests.exceptions.HTTPError as error:
+            response = error.response
+            status_code = response.status_code
+            message = response.json().get('message')
+
+            if status_code == requests.codes.bad_request and message is not None:
+              raise Exception(message)
+            else:
+              raise error
+
```
rlmv: You shouldn't need any of this error handling - the way that the client wraps requests means that any non-200s are already raised.

blackfynn/__init__.py
```diff
@@ -28,4 +28,4 @@
 )
 
 __title__ = 'blackfynn'
-__version__ = '2.8.0'
+__version__ = '2.8.1'
```
rlmv: Let's go to `2.9.0` since this is adding a new feature.

blackfynn/api/data.py
rohanshah: eh not really, here is what it looks like with this error handling:
```
In [3]: p.process()
---------------------------------------------------------------------------
Exception                                 Traceback (most recent call last)
<ipython-input-3-e73e4609c254> in <module>()
----> 1 p.process()

/Users/rohan/blackfynn/code/blackfynn-python/blackfynn/models.py in process(self)
    807         """
    808         self._check_exists()
--> 809         return self._api.packages.process(self)
    810
    811     def relate_to(self, *records):

/Users/rohan/blackfynn/code/blackfynn-python/blackfynn/api/data.py in process(self, pkg)
    241
    242             if status_code == requests.codes.bad_request and message is not None:
--> 243               raise Exception(message)
    244             else:
    245                 raise error

Exception: Can only process a package in the UPLOADED state.
```

and here is what it looks like without:
```
In [2]: p.process()
---------------------------------------------------------------------------
HTTPError                                 Traceback (most recent call last)
<ipython-input-2-e73e4609c254> in <module>()
----> 1 p.process()

/Users/rohan/blackfynn/code/blackfynn-python/blackfynn/models.pyc in process(self)
    807         """
    808         self._check_exists()
--> 809         return self._api.packages.process(self)
    810
    811     def relate_to(self, *records):

/Users/rohan/blackfynn/code/blackfynn-python/blackfynn/api/data.py in process(self, pkg)
    243             #  raise Exception(message)
    244             #else:
--> 245             raise error
    246
    247     def get_sources(self, pkg):

HTTPError: 400 Client Error: Bad Request for url: https://dev.blackfynn.io/packages/N%3Apackage%3A877382f0-ac6d-49ce-9cdb-48c70aa705f1/process
```

i.e. no description of what happened

rlmv: Word, that is a lot nicer. 👍

rohanshah:
@rlmv in response to your top-level questions:
As per the JIRA ticket (which I just added to the description, sorry, should have been there initially) I don't think we want to add the API calls to toggle that settings from the python client.

For the second, I can't really right now, except for the error-case if you want. The code that allows an upload to remain in the UPLOADED state isn't currently out. I can make a ticket to add a success test later when it is available?

rlmv:
Gotcha, makes sense. Let's keep the PR open until then if that's ok? It'd be great to keep master in a deployable state since we have some other Python changes hopefully going out early next week.

rohanshah:
@rlmv no problem leaving it open, however just want to note, it is in a deployable state, even if the endpoint/function will always error right now (it should because technically all datasets automatically process packages so no package should be able to call process on itself i.e. this is desired now and going forward, technically)

rlmv:
Gotcha 👍

blackfynn/models.py
```diff
@@ -2100,6 +2109,8 @@ def as_dict(self):
         return dict(
             name = self.name,
             description = self.description,
+            status = self.status,
+            automaticallyProcessPackages = self.automatically_process_packages,
```
rlmv: I think these should not be included in `as_dict`. My understanding is that we don't want either of these to be editable from the Python client - is that true?

blackfynn/client.py
```diff
@@ -185,7 +185,7 @@ def create(self, thing):
         """
         return self._api.core.create(thing)
 
-    def create_dataset(self, name):
+    def create_dataset(self, name, description=None, automatically_process_packages=False):
```
rlmv: This is going to be breaking change for users creating a dataset. Can you update `docs/tutorials/data_catalog.rst` to describe how and why to toggle this feature?

tests/test_io.py
```diff
@@ -41,6 +43,29 @@ def test_upload(client, dataset, upload_args, n_files):
...
+    package = collection.items[0]
+
+    # wait for max of one minute or until the package is in the UPLOADED state
+    wait_until = time.time() + 60
+    while time.time() < wait_until and package.state != "UPLOADED":
+        time.sleep(5)
+        package = client.get(package.id)
+
+    assert package.state == "UPLOADED"
+    assert package.process()
```
rlmv: Yikes, yeah this isn't great. Will `package.process()` fail if the package is not in `UPLOADED`?

rohanshah: yes (unless it's TimeSeries and in the `READY` state and an append job is happening but that's a similar problem)

blackfynn/client.py
rohanshah: why is it breaking if there are defaults? or do you mean the functionality will be different?

rlmv: In that the functionality will be different. I think that's ok since this is a policy change, but let's be sure to make that clear in the documentation.

blackfynn/models.py
rohanshah: Not sure what you mean by editable. `automatically_process_packages` can be set via the `create_dataset` function. That function requires `as_dict` with automatically processing in it.

I can remove `status` though

blackfynn/client.py
rohanshah: I think breaking implies that if you upgrade to this version your code will either break (i.e. not run) or behave differently than before. But neither are true. If you upgrade or not, calling this function in that way you had previously will work identically.

rohanshah: Will add to the docs though.

blackfynn/models.py
rlmv: Updateable, changable. I didn't realize this was used for creating datasets so makes sense that we need it. Let's do get rid of status though

blackfynn/client.py
rlmv: Behavioral changes are also breaking changes - more subtle, but still breaking, and could affect user workflows.

rlmv: Oh dang, I'm sorry -  I was reading `automatically_process_packages=False` to be `True`. Forget I said anything.

rlmv: The previous behavior was `automatically_process_packages = True`

rlmv: But again, I'm not opposed to it - just want to call that out in the docs.

rohanshah: That is true but it's not a behavioral change to *the client*. It's a behavioral change to our APIs and platform (which we don't version as of right now). Regardless of whether a user uses client version `2.8.0` and `2.9.0` `automatically_process_packages =  True` as of now and `automatically_process_packages =  False` as of the *API* release, not the client release. In fact if this is never released `automatically_process_packages` default will change from `True` to `False`.  It's the client that is versioned and documented here, not the API and our platform at large.

rlmv: Good point, good point. I didn't think about how the old clients are going to behave with this change.

rlmv: 👍

rohanshah: > The previous behavior was `automatically_process_packages = True`, so this is a change.

Previous to what though? This client release? If so, that's not true. I can never release this client and next week it will now be the case that the previous version (to this client) was `automatically_process_packages = False`. Would it still be considered a breaking change?

rohanshah: ah sorry keep writing responses after you've already written back haha wish github auto-updated comment threads

rohanshah: also, added docs, let me know if it was what you were looking for or something else!

rlmv: Hah, it's all good.

tests/test_io.py
rohanshah: yeah ... this isn't really great but there isn't really a way around it. I don't mind having a test for coverage that ensures this function can be called and error with a `400` and not a `404` or something but it wouldn't be a very functional test.

rlmv:
This all looks good to me. We're going to have an integration with the agent going out as well, let's coordinate and get those out under the same version.

rohanshah:
@rlmv sounds good to me, note that @natevecc is working with Matt to add this same functionality to the Rust client and agent so that works out perfectly.

blackfynn/models.py
```diff
@@ -2100,6 +2109,7 @@ def as_dict(self):
         return dict(
             name = self.name,
             description = self.description,
+            automaticallyProcessPackages = self.automatically_process_packages,
```
rlmv: Sorry to come back with one more thing: could we change the `automatically_process_packages` attribute to just be `process_packages`? `automatically_process_packages` is pretty verbose compared to most of the other properties and instance variables in the client, and we're already mapping the name in `as_dict` so should be as simple as adding a `from_dict` method that looks like

```
@classmethod
      data['process_packages'] = data.get('automaticallyProcessPackages', False)
      return super(Dataset, cls).from_dict(data, *args, **kwargs)
```
(Last request, promise. Also not a hard need, but think it would help keep things tidy.)

rohanshah: Yeah I thought about that and agree that it's overly verbose.

This reason I kept it is that I wanted to maintain consistency with the API, and also, personally, i found it confusing to refer to it as just `process_packages` because it seems like if you set that to `False`, packages won't (ever?) be processed (even though you can manually) rather than just not _automatically_ processed.

I'm not hard on this either, it's just how I felt after having to name it on the API side as well.

rlmv: Hmm. `always_process_packages`? @merky @zduey do you have any naming preferences here?

zduey: I agree it is verbose, but I think `automatically_process_packages` is the least of the evils. I prefer the consistency with the API and to minimize manually mapping between different names in different code bases as much as possible.

merky: My opinion is that these should really be called something like `generateViewerAssets` or `createViewerAssets` or something else along those lines, because `processPackage` is not what's going on — this is very specific to generating data that will enable web viewers, and I don't think that's been clear in the code elsewhere. Start here!? Tack on an `auto` if you want :)

rohanshah: @merky that's a good point actually but do the terms "viewer" or "asset" really appear anywhere? I hesitate to name anything that would _add_ confusion.

rohanshah: also I don't think it's just for viewers, it also makes some data programmatically accessible like streaming timeseries data.

PR #146, opened 2019-03-14
Workspace querying updates

zduey:
# Description

Adds support for querying snapshots in a workspace

## Type of change

# How Has This Been Tested?

# Checklist:


blackfynn/models/workspace.py
```diff
@@ -139,6 +143,46 @@ def get_all_named_queries(self):
...
+    def get_query_results_presigned_url(self, execution_id):
+        """ Get download link for the query results
+        Args:
+            execution_id (str): Query execution ID
+        Returns:
+            str
+        """
+        return self._api.analytics.get_query_results_presigned_url(
+            self, execution_id)
+
```
rlmv: I want to simplify this API. Could something like the following work?

- `self._api.analytics.execute_query` returns the `QueryResults` object or a wrapper that just tracks an `execution_id`
- The `QueryResults` then has a `status` property/method (that calls `get_query_execution_status`) and the `download` and `as_dataframe` methods. These should raise an exception if the results are not ready. (I'm not sure if this requires you to get the `status` first, or if the results endpoint gives back that information.)

This would make it so that `NamedQuery.execute` and `Workspace.execute_query` return the same objects and make the user flow look something like this:

```
results = named_query.execute()
results = workspace.execute_query("SELECT * ...")  

results.status                # Dynamically retrieved
results.download()        # exception if not available
results.as_dataframe()  # exception if not available
```

I think this would be a lot cleaner than having users know about `execution_id`.

zduey: Yep, that is a cleaner API from the user perspective. I was trying to keep the extra work that the python client is doing to a minimum so that it is more of a basic client to interact with the API as it is defined, which requires working with the `execution_id`. Requiring the user to keep track of the execution id is also how the API for the Athena client works. I guess this is more of a philosophical difference as to what a client should be doing. I prefer the minimalist approach so that the language-specific clients are a simple wrapper over the actual API. That said, offering a more intuitive interface is more consistent how the rest of the python client works.

Some day, I'd like to take the AWS approach and have two ways to interact with our APIs: one low-level client interface that is this really thin wrapper around the API and then a higher-level interface for interacting with resources on the platform via language-specific objects. It would make it easier to auto-generate the lower-level APIs based on swagger or some other IDL. I know this separation exists in the python client today, we just don't give users explicit access to the lower-level api.

tests/test_analytics.py
```diff
@@ -55,13 +59,32 @@ def ready_snapshot(graph_view):
...
     return graph_view.latest()
 
+
+@pytest.fixture(scope='module')
+def ready_query_results(client, workspace, ready_snapshot):
+    table_name = ready_snapshot.view.name
+    query_results = workspace.execute_query(
+        'SELECT * FROM "{}"'.format(table_name))
+    start = time.time()
+    while client._api.analytics.get_query_execution_status(workspace, query_results.execution_id) is not 'succeeded' and (time.time() - start < 30):
```
rlmv: `while query_results.status is not 'succeeded' and (time.time() - start < 30):` ??

tests/test_analytics.py
```diff
@@ -197,27 +236,90 @@ def test_delete_contents(client, simple_graph):
...
+
+def test_execute_named_query(workspace, ready_snapshot):
+    table_name = ready_snapshot.view.name
+    query = workspace.create_named_query(
+        "Live", 'SELECT * FROM "{}"'.format(table_name))
+    results = query.execute()
+    assert results.workspace_id == workspace.id
+
+
+def test_get_query_execution_status(client, ready_query_results):
```
rlmv: Could this test be removed? It tests the same thing as `test_query_results_status` but at a lower level of abstraction.

tests/test_analytics.py
```diff
@@ -197,27 +236,90 @@ def test_delete_contents(client, simple_graph):
...
+    status = client._api.get_query_execution_status(ready_query_results)
+    assert status is not None
+
+
+def test_query_results_status(ready_query_results):
+    assert ready_query_results.status == "succeeded"
+
+
+def test_get_query_execution_results(client, ready_query_execution_id):
```
rlmv: And this one?

tests/test_analytics.py
zduey: Yeah, that's more concise. I was trying to use the api method so I wasn't also testing the status property when creating the fixture, but it really doesn't matter.

tests/test_analytics.py
zduey: 💣

tests/test_analytics.py
zduey: 💣

tests/test_analytics.py
rlmv: 👍I like using the highest level abstraction that gets the job done so we can change the implementation without cascading edits everywhere.

PR #152, opened 2019-04-08
Add JWT support

mikeswoods:
- Add JWT support as an alternative authorization method to the existing API/secret key method

https://blackfynn.atlassian.net/browse/AP-936

## Type of change

# How Has This Been Tested?

Manual testing was conducted. Automated testing is difficult as there are no public methods in API to generate a JWT (which makes sense as they are meant to be internal). Currently, I can use `auth-middleware-utilities` (soon to be rolled into `auth-middleware`) to generate a JWT for testing.

In `test_jwt.py`:
```bash
from blackfynn import Blackfynn
import os

assert os.getenv('BLACKFYNN_JWT') is not None

bf = Blackfynn(host="https://dev-api-use1.blackfynn.io", concepts_host="https://concepts.dev.blackfynn.io")
for ds in bf.datasets():
    print(ds)
```
and

```bash
$ BLACKFYNN_JWT=<generated-token> python test_jwt.py
```

# Checklist:


blackfynn/base.py
```diff
@@ -156,10 +180,16 @@ def _call(self, method, endpoint, base='', *args, **kwargs):
         else:
             host = self._host
 
+        if 'reauthenticate' in kwargs:
+            reauthenticate = kwargs['reauthenticate']
+            kwargs.pop('reauthenticate')
+        else:
+            reauthenticate = True
```
rlmv: Instead of this block, you can just add `reauthenticate=True` as a new keyword argument to `_call`

blackfynn/base.py
```diff
@@ -180,15 +210,16 @@ def _put(self, endpoint, *args, **kwargs):
...
         try:
             return req.call(timeout=self.settings.max_request_time)
         except UnauthorizedException as e:
             if self._token is None:
                 raise e
 
             # try to refresh the session and re-request
-            self.authenticate(self._organization)
+            if reauthenticate:
+                self.authenticate(self._organization)
```
rlmv: If `reauthenticate` is `False` we should not try to call `req.call` again. I think this except block should be

```
        except UnauthorizedException as e:
            if self._token is None or reauthenticate is False:
                raise e

            # try to refresh the session and re-request
             self.authenticate(self._organization)
             return req.call(timeout=self.settings.max_request_time)
```

blackfynn/client.py
```diff
@@ -86,21 +86,24 @@ class Blackfynn(object):
...
             'concepts_api_host': concepts_host,
             }.items() if v != None })
         self.settings = Settings(profile, overrides, env_override)
 
-        if self.settings.api_token  is None: raise Exception('Error: No API token found. Cannot connect to Blackfynn.')
-        if self.settings.api_secret is None: raise Exception('Error: No API secret found. Cannot connect to Blackfynn.')
+        if (self.settings.api_token is None or self.settings.api_secret is None) and self.settings.jwt is None:
+            if self.settings.api_token  is None: raise Exception('Error: No API token found. Cannot connect to Blackfynn.')
+            if self.settings.api_secret is None: raise Exception('Error: No API secret found. Cannot connect to Blackfynn.')
+            if self.settings.jwt  is None: raise Exception('Error: No JWT found. Cannot connect to Blackfynn.')
```
rlmv: I know this was like this before, but can you move the `raise Exception` bodies of the if-statements to a new line? I think it will read easier.

blackfynn/base.py
mikeswoods: Good call

rlmv:
On another note, it looks like the tests are failing because of the Agent - I'm seeing `error: missing field `fileType` at line 1 column 360` when running the agent tests. Is that a known problem?

blackfynn/base.py
```diff
@@ -70,9 +71,17 @@ def __init__(self, settings):
...
+        """
+        if self._jwt is None:
+            self._authenticate_with_session(organization = organization)
+        else:
+            self._authenticate_with_jwt(organization = organization)
+
+    def _authenticate_with_session(self, organization = None):
+        """
+        An API token is used to authenticate against the Blackfynn platform.
+        The token that is returned from the API call will be is used for all
```
mattusifer: "will be" vs "is"

PR #154, opened 2019-04-09
AP-975: Expose query/run endpoint in blackfynn-python

mikeswoods:
# Description

This PR exposes the concepts service `query/run` endpoint.

## JIRA Issue

https://blackfynn.atlassian.net/browse/AP-975

## Type of change

# How Has This Been Tested?

Added a simple graph filtering test.

# Checklist:


blackfynn/api/concepts.py
```diff
@@ -524,6 +529,195 @@ def create(self, dataset, external_id, relationship, concept_instance, values, d
...
+            "joins": [j.as_dict() for j in self._joins],
+            "offset": self._offset,
+            "limit": self._limit,
+            "orderBy": { "Ascending": { "field": "$createdAt" } },
+        }
+        if self._select is not None:
+            query['select'] = self._select.as_dict()
+        return query
+
+    def query_json(self):
```
rlmv: This method is unused.

blackfynn/api/concepts.py
```diff
@@ -524,6 +529,195 @@ def create(self, dataset, external_id, relationship, concept_instance, values, d
...
+            Review.query() \
+                .filter("is_complete", "eq", False) \
+                .run()
+        """
+        assert isinstance(key, str), "key must be a string"
+        assert operator in self.allowed_operators, "not a valid predicate operator: {}".format(operator)
+        self._filters.append(ModelFilter(key, operator, value))
+        return self
+
+    def join(self, target, *filters):
```
rlmv: Not a hard requirement, but it would be cool if you could pass a `Model` instance as a target in addition to a string, both here and in `select`.

tests/test_concepts.py
```diff
@@ -456,3 +456,12 @@ def test_get_graph_summary(simple_graph):
...
     assert summary['relationshipTypeCount'] == 1
+
+def test_simple_query(simple_graph):
+    model_a = simple_graph.dataset.get_model('Model_A')
+    # Filter
+    assert len(model_a.query().filter("prop1", "eq", "val1").run()) == 1
+    assert len(model_a.query().filter("prop1", "eq", "not-present").run()) == 0
+    # Join
+    assert len(model_a.query().join("Model_B", ("prop1", "eq", "val1")).run()) == 1
+    assert len(model_a.query().join("Model_B", ("prop1", "eq", "val2")).run()) == 0
```
mattusifer: It would be great to get a few assertions around the other operations - `select`, `offset` and `limit`. Maybe an assertion or two that strings multiple operations together in any order.

blackfynn/api/concepts.py
mikeswoods: good idea

blackfynn/api/concepts.py
mikeswoods: removed

tests/test_concepts.py
mikeswoods: good call

PR #159, opened 2019-06-20
Update package fix

JeremyLecoeur:
# Description

Currently, the python library allow user to mess with the normal flow of state from the state machine.
We propose removing the state from the update function and also removing all the functions that mutate the state

## Github Issue

https://app.clickup.com/t/11h2q5

## Type of change

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [X] This change requires a documentation update

# Checklist:

### DIFF
--- a/blackfynn/models.py
+++ b/blackfynn/models.py
@@ -448,29 +448,6 @@ class BaseDataNode(BaseNode):
         r = self._api.core.delete(self)
         self.id = None
 
-  #  def set_ready(self, **kwargs):
-  #      """
-  #      Set's the package's state to ``READY``
-  #      """
-  #      self.state = "READY"
-  #      return self.update(**kwargs)
-
-  #  def set_unavailable(self):
-  #      """
-  #      Set's the package's state to ``UNAVAILABLE``
-  #      """
-  #      self._check_exists()
-  #      self.state = "UNAVAILABLE"
-  #      return self.update()
-
-  #  def set_error(self):
-  #      """
-  #      Set's the package's state to ``ERROR``
-  #      """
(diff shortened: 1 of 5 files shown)


blackfynn/models.py
```diff
@@ -448,28 +448,28 @@ def delete(self):
...
-        return self.update()
-
-    def set_error(self):
-        """
-        Set's the package's state to ``ERROR``
-        """
-        self._check_exists()
-        self.state = "ERROR"
-        return self.update()
+  #  def set_ready(self, **kwargs):
```
mattusifer: since these are technically public facing maybe we should deprecate these for now instead? then remove them in the next release.

mattusifer: we spoke offline and decided it was best to just remove this functionality, it is detrimental to users.

JeremyLecoeur: I removed the functions all together rather than commented them out

docs/models.rst
```diff
@@ -60,7 +60,7 @@ The ``BaseDataNode`` class provides the basic methods available on **all models*
 .. autoclass:: BaseDataNode
    :members:
    :inherited-members:
-   :exclude-members: add_properties, insert_property
+   :exclude-members: add_properties, insert_property, set_ready, set_error, set_unavailable
```
rlmv: Since they are being removed, these methods don't need to be in the Sphinx docs.

rlmv:
Will this be released as version 2.13.0?

docs/models.rst
JeremyLecoeur: good point

PR #162, opened 2019-07-03
Add support for linked properties

technoto:
# Description

I implemented all of the linked property API endpoints. This includes:
- Get all linked properties in a dataset.
- Get/add/update/delete linked properties from a model.
- Get/add/delete linked property values from a record.

I also created front-facing classes `LinkedModelProperty` and `LinkedModelValue` to handle API calls.

## Type of change

# How Has This Been Tested?

I tested against all existing unit tests, and included tests to make sure each API method works as expected.

# Checklist:


blackfynn/models.py
```diff
@@ -2162,6 +2162,22 @@ def as_dict(self):
...
+    def get_linked_property(self, prop):
+        """
+        Get a LinkedModelProperty by name or id.
+        """
+        all_props = self.get_linked_properties()
+        for p in all_props:
+            if prop == p.name or prop == p.id:
+                return p
+        raise Exception("No linked property found with name or id '{}'".format(prop))
+
```
rlmv: I think we should remove both of these methods from the `Dataset` object, since linked properties are specifically scoped to models. Additionally, linked property names do not have to be unique across datasets, so calling `get_linked_property` with a name will arbitrarily return a single linked property when multiple may exist. For example, the following is valid:
```
a = dataset.create_model('a')
b = dataset.create_model('b')
c = dataset.create_model('c')

lpa = a.add_linked_property('link_to_c', c)
lpb = b.add_linked_property('link_to_c', c)

dataset.get_linked_property('link_to_c') # Could return lpa OR lpb

```
The corresponding implementations on `Model` are great though.

blackfynn/models.py
```diff
@@ -3283,6 +3434,49 @@ def relate_to(self, destinations, relationship_type='related_to', values=None, d
         # use batch endpoint to create relationships
         return self._api.concepts.relationships.instances.create_many(self.dataset_id, relationship_type, *relationships)
 
+    def get_links(self):
```
rlmv: I wonder if we should name `get_links` and `get_link` something that describes what is being linked to, eg `get_linked_record`? Do you think that would be easier for a user to grasp?

tests/test_linked.py
```diff
@@ -0,0 +1,160 @@
...
+    DataPackage,
+    LinkedModelProperty,
+    LinkedModelValue,
+    ModelProperty,
+    ModelPropertyEnumType,
+    ModelPropertyType,
+    convert_datatype_to_type,
+    convert_type_to_datatype,
+    uncast_value
+)
```
rlmv: This test suite is a thing of beauty. Thank you 🙏

tests/test_linked.py
```diff
@@ -0,0 +1,160 @@
...
+from tests.utils import create_test_dataset, current_ts
+
+### Create a fresh dataset for each test:
+@pytest.fixture(scope='function')
+def dataset(client):
```
rlmv: So that we're not overriding the existing `dataset` fixture, how would you feel about renaming this to something like `empty_dataset`?

blackfynn/models.py
```diff
@@ -3283,6 +3434,49 @@ def relate_to(self, destinations, relationship_type='related_to', values=None, d
...
+        """
+        Get a link value by name or id.
+        """
+        all_links = self.get_links()
+        for l in all_links:
+            if link == l.name or link == l.id:
+                return l
+        raise Exception("No link found with a name or ID matching '{}'".format(link))
+
+    def link_to(self, target, link):
```
rlmv: There is an interesting corner case here: if you have a linked property value on a record (let's call it `lp1`) and you add a second value (`lp2`) for the same linked property, API does not remove the old linked record. It will still exist in the concepts service, but will not be accessible. However, if you then delete `lp2`, `lp1` will appear again. This is a bug in the backend that really needs to be fixed. Until then, can you update this method to delete any existing links before adding a new one? I think the frontend has a similar workaround.

See https://app.clickup.com/t/y55gn for the ticket.

blackfynn/models.py
```diff
@@ -2454,6 +2454,57 @@ def set_type(self, type):
...
+        self.position = position
+
+        if display_name is None:
+            self.display_name = name
+        else:
+            self.display_name = display_name
+
+        if isinstance(target, Model):
+            self.target = target.id
+        elif isinstance(target, str):
```
rlmv: This probably needs to be updated to use `string_types` instead of `str` for Python 2 compatibility.

PR #163, opened 2019-07-08
Allow model values to be arrays

rlmv:
# Description

Refactor the internals of model property types to allow array values to be correctly set. 

This change moves all the logic of validating data types and casting property values into the `ModelPropertyType` and `ModelPropertyEnumType` objects. This makes it easier to special-case situations that are only applicable to enumerated types. In addition, internal types that were previously represented by Python literals (`str`, `int`, etc) are uniformly converted to `ModelPropertyType` instances, making it easier to reason about the code.

With this structure it would be easy to write separate property type classes for each datatype, eg `ModelPropertyStringType`, `ModelPropertyDate` type, removing most of the `isinstance` checks that are needed for special cases (around dates particularly.)

This PR also deprecates the `ModelProperty.set_type` and `ModelValue.set_value` methods, replacing them with Python `setter` attributes.

## Type of change

# How Has This Been Tested?

# Checklist:


blackfynn/models.py
```diff
@@ -2283,72 +2216,178 @@ def target_type_string(target):
...
+    @property
+    def _blackfynn_type(self):
+        return python_to_blackfynn_type_map[self.data_type]
+
+    @staticmethod
+    def from_dict(data):
+        if isinstance(data, ModelPropertyType):
+            return data
+
+        elif isinstance(data, dict):
```
zduey: this is so close to being able to be a classmethod. Could you pull out this little bit of logic to differentiate between `ModelPropertyEnumType`s and `ModelPropertyType`s into a different static method so that the `from_dict` methods on the two objects can stay classmethods?

blackfynn/models.py
```diff
@@ -2447,39 +2494,48 @@ def enum(self):
...
         return u"<BaseModelProperty name='{}' {}>".format(self.name, self.type)
 
+
 class BaseModelValue(object):
-    def __init__(self, name, value, *args, **kwargs):
+    def __init__(self, name, value, data_type=None):
         assert " " not in name, "name cannot contain spaces, alternative names include {} and {}".format(name.replace(" ", "_"), name.replace(" ", "-"))
 
         self.name = name
+        self.data_type = ModelPropertyType.from_dict(data_type)
```
zduey: I think this would have to change to use the other dispatcher utility if you made `from_dict` a classmethod

blackfynn/models.py
rlmv: Good idea.

blackfynn/models.py
```diff
@@ -2283,72 +2216,178 @@ def target_type_string(target):
...
         super(ModelPropertyEnumType, self).__init__(data_type, format, unit)
+
+        if enum is not None:
+            enum = list(enum)
         self.enum = enum
+
         self.multi_select = multi_select
         self.selection_type = 'array' if self.multi_select else 'enum'
 
+    def _decode_value(self, value):
```
technoto: One small change - we should check if `value` is None or else it'll throw an exception if you try to iterate over it. Same thing with `_encode_value()`.

rlmv: Good call - we do that for the non-array case already.

PR #167, opened 2019-07-16
#u2y7b inject response body api-related errors raised by the python client

JeremyLecoeur:
# Description

`resp.raise_for_status()` does not inject the response body. We use this method when re-raising errors in the python client. Without the response body, the helpful information provided by the API is lost. 

## Github Issue

https://app.clickup.com/t/u2y7b

## Type of change

# How Has This Been Tested?

Manual tests + adding test in test suite

# Checklist:

### DIFF
--- a/blackfynn/base.py
+++ b/blackfynn/base.py
@@ -8,6 +8,7 @@ import json
 import requests
 from requests import Session
 from requests.adapters import HTTPAdapter
+from requests.exceptions import HTTPError
 from requests.packages.urllib3.util.retry import Retry
 
 # blackfynn
@@ -28,6 +29,17 @@ class BlackfynnRequest(object):
         self._response = None
 
         self._logger = log.get_logger('blackfynn.base.BlackfynnRequest')
+        
+    def expanded_raise_for_status(self,resp):
+        max_text_length=500
+        try:
+            resp.raise_for_status()
+        except HTTPError as e: #raise for status raise an HTTPError, so we can use it to grab the message
+            if resp.text:
+                raise HTTPError(u' Responsse Body: %s' % resp.content[:max_text_length], response=resp)
+            else:
+                raise e
+        return
 
     def _handle_response(self, resp):
         self._logger.debug(u"resp = {}".format(resp))
@@ -36,7 +48,7 @@ class BlackfynnRequest(object):
             raise UnauthorizedException()
 
         if not resp.status_code in [requests.codes.ok, requests.codes.created]:
-            resp.raise_for_status()
+            self.expanded_raise_for_status(resp)
         try:
             # return object from json
             resp.data = json.loads(resp.text)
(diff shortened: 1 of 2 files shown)


blackfynn/base.py
```diff
@@ -28,6 +29,17 @@ def __init__(self, func, uri, *args, **kwargs):
...
 
         self._logger = log.get_logger('blackfynn.base.BlackfynnRequest')
+        
+    def expanded_raise_for_status(self,resp):
+        max_text_length=500
+        try:
+            resp.raise_for_status()
+        except HTTPError as e: #raise for status raise an HTTPError, so we can use it to grab the message
+            if resp.text:
+                raise HTTPError(u' Responsse Body: %s' % resp.content[:max_text_length], response=resp)
```
rlmv: I think this should use the `raise_from` function - it will wrap the original exception in the first one and show the full stack trace and error message of both. See https://github.com/Blackfynn/blackfynn-python/blob/499294159d5d5aeca52d9707f68de81d8c405782/blackfynn/api/agent.py#L54 for an example.

How do you feel about removing the "Response Body:" prefix from the error message? I think there's enough context with the `HttpError` itself.

I also think we should display the full error message instead of truncating it to 500 characters. Especially in dev, we often see stack traces that are much longer than that.

blackfynn/base.py
```diff
@@ -28,6 +29,17 @@ def __init__(self, func, uri, *args, **kwargs):
         self._response = None
 
         self._logger = log.get_logger('blackfynn.base.BlackfynnRequest')
+        
+    def expanded_raise_for_status(self,resp):
```
rlmv: This could just be called `raise_for_status` since the Requests version isn't in scope here.

JeremyLecoeur: ok

blackfynn/base.py
JeremyLecoeur: done!

blackfynn/base.py
```diff
@@ -28,6 +30,17 @@ def __init__(self, func, uri, *args, **kwargs):
         self._response = None
 
         self._logger = log.get_logger('blackfynn.base.BlackfynnRequest')
+        
+    def raise_for_status(self,resp):
+        max_text_length=500
```
rlmv: This variable is now unused.

blackfynn/base.py
```diff
@@ -28,6 +30,17 @@ def __init__(self, func, uri, *args, **kwargs):
...
+        try:
+            resp.raise_for_status()
+        except HTTPError as e: #raise for status raise an HTTPError, so we can use it to grab the message
+            if resp.text:
+                raise_from(HTTPError(resp.content,response=resp), e)
```
rlmv: Whitespace after the `,` please.

blackfynn/base.py
JeremyLecoeur: indeed

PR #172, opened 2019-07-23
Remove (most of) the CLI and other deprecated code

technoto:
# Description

This PR removes a lot of deprecated code from the Python client:

**Python CLI**
- renamed from `bf` to `bf_profile`
- removed all functionality except for connection status and profile management
- **NOTE:** The user must rebuild this project in order to reflect these changes.

**Python API**
- removed several deprecated methods
- set `use_agent=True` by default for uploading files

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

# How Has This Been Tested?
- Tested against existing unit tests
- Updated `test_io.py` to make sure the agent is used by default when uploading
- Manually tested all of the remaining CLI commands

# Checklist:


blackfynn/cli/bf.py
```diff
@@ -1,148 +1,289 @@
...
 from docopt import docopt
 
 import blackfynn
-from blackfynn import Blackfynn, Settings
-
-from .working_dataset import set_working_dataset
+from blackfynn import Blackfynn, DEFAULT_SETTINGS, Settings
 
 
 DEPRECATION = """
```
rlmv: I think we can get rid of this warning, and maybe just add something like `For additional commands, please install the Blackfynn CLI: https://developer.blackfynn.io/agent` at the bottom of the main help docstring.

blackfynn/cli/bf.py
```diff
@@ -1,148 +1,289 @@
...
-
-  props             Add/remove/modify a package or collection's properties
-  move              Move a package or collection
-  where             Show path to package or collection
-  append            Append data to a package
-  rename            Rename a package or collection
-  delete            Delete a package or collection
-  create            Create a collection
-  get               Get the contents of a package, collection, or dataset
+'''
```
rlmv: Could this file be renamed `bf_profile` to match how it will be invoked?

blackfynn/cli/bf_status.py
```diff
@@ -4,15 +4,11 @@
```
rlmv: This is now a pretty small bit of code - do you think it would be clearer if it was just a function in the `bf_profile` file?

blackfynn/models.py
```diff
@@ -1916,37 +1915,6 @@ def __init__(self, name, description=None, status=None, automatically_process_pa
     def __repr__(self):
         return u"<Dataset name='{}' id='{}'>".format(self.name, self.id)
 
-    @property
```
rlmv: Could you remove the corresponding endpoints/methods in `api/data.py` as well?

tests/test_io.py
```diff
@@ -23,45 +22,43 @@ def _resource_path(fname):
 INNER_DIR  = 'inner_dir'
 
 
-@pytest.mark.agent
-def test_upload_with_agent_to_dataset(client, dataset):
+@pytest.mark.legacy
```
rlmv: There's some pytest magic in the top-level `conftest.py` that skips all the `agent` tests if the user doesn't pass the `--agent` flag. I think it makes sense to keep the `agent` marker and just change the pytest command-line flag to be `--skip-agent`. Does that make sense? You can also remove the usage of `--agent` from the `travis.yml` build file. We also need to update the `Dockerfile` that is used as an end-to-end test for this project to either install the agent, or skip the agent tests.

technoto: Hmm, I don't see anything related to `--agent` in `conftest`.

Dockerfile
```diff
@@ -5,6 +5,11 @@ ADD requirements-test.txt /app/requirements-test.txt
 RUN pip install -r /app/requirements.txt
 RUN pip install -r /app/requirements-test.txt
 
+# install agent:
+ENV AGENT_VERSION="0.2.6"
+RUN wget "http://data.blackfynn.io.s3.amazonaws.com/public-downloads/agent/${AGENT_VERSION}/x86_64-unknown-linux-gnu/blackfynn-agent_${AGENT_VERSION}_amd64.deb" -O agent.deb
+RUN sudo dpkg -i agent.deb
+
```
technoto: Not totally familiar with Docker so I'm not sure if this is right? But I basically copied the installation steps in `.travis.yml`.

rlmv: Yep, this looks right. In order to make better use of Dockers layer cache, I think these should probably be the first commands after `FROM python:2.7`. 

If `docker build .` works you should be good to go.

conftest.py
```diff
@@ -1,15 +1,16 @@
...
 dependent tests.
 """
+import pytest
+
 
 def pytest_addoption(parser):
-    parser.addoption("--agent", action="store_true", help="run agent integration tests")
+    parser.addoption("--skip-agent", action="store_true", help="run agent integration tests")
 
+def pytest_configure(config):
```
rlmv: 👍

docs/quickstart.rst
```diff
@@ -12,6 +12,8 @@ The Python client is compatible with Python 2.7 and 3.4-3.7.
 
     $ pip install -U blackfynn
 
+The **Blackfynn Agent** is installed separately and is required for optimal performance when uploading files.
```
rlmv: 👍

docs/tutorials/data_catalog.rst
```diff
@@ -47,10 +47,6 @@ In this example, the only datasets available are ``Test Data`` and ``My Research
 Creating new data
 ^^^^^^^^^^^^^^^^^^
 
-.. note::
```
rlmv: Could we keep this note here in a simpler form, eg `See :ref:`agent` for more information on uploading data.`?

setup.py
```diff
@@ -36,7 +36,7 @@
     python_requires='>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, <4.0',
     entry_points = {
         'console_scripts': [
-            'bf=blackfynn.cli.bf:blackfynn_cli',
+            'bf_profile=blackfynn.cli.bf_profile:main',
```
rlmv: 👍

blackfynn/cli/bf_profile.py
```diff
@@ -223,20 +239,54 @@ def list_keys(settings):
...
+    print("Blackfynn environment:")
+    print("  User               : {}".format(bf.profile.email))
+    print("  Organization       : {} (id: {})".format(bf.context.name, bf.context.id))
+    print("  API Location       : {}".format(bf.settings.api_host))
+    print("  Streaming API      : {}".format(bf.settings.streaming_api_host))
+
 
 #Helper functions
 #=======================================================
+def yesno_prompt(msg):
```
rlmv: 👍 Nice cleanup

PR #173, opened 2019-08-21
#1pa3nt: Refactor profile management script

rlmv:
# Description

Cleans up the profile management script:

- Leverage `docopt` to extract arguments instead of handling them manually
- Prefer an early-return coding style to reduce the number of nested `if`-statements
- Exit with non-zero return codes when commands fail or are invalid
- Create a new `unset-default` command instead of passing `'none'` to the `set-default` command

## Type of change

# How Has This Been Tested?

Tested all commands manually. It would be great to write a test suite for this as well...

# Checklist:


blackfynn/cli/bf_profile.py
```diff
@@ -48,40 +53,48 @@
...
-    elif args['<command>'] == 'set-default' : set_default(settings, args['<args>'])
-    elif args['<command>'] == 'list'        : list_profiles(settings, args['--contents'])
-    elif args['<command>'] == 'show'        : show_profile(settings, args['<args>'])
-
-    elif args['<command>'] == 'set'         : set_key(settings, args['<args>'], args['--force'])
-    elif args['<command>'] == 'unset'       : unset_key(settings, args['<args>'], args['--force'])
-    elif args['<command>'] == 'keys'        : list_keys(settings)
-
-    elif args['<command>'] == 'status':
+    elif args['create']:
```
mikeswoods: Much clearer

blackfynn/cli/bf_profile.py
```diff
@@ -166,79 +177,97 @@ def list_profiles(settings, contents):
...
+        name = input('  Profile name to set as default: ')
 
-    if name == 'none':
-        print("Default profile unset. Using global settings and environment variables")
-        settings.config['global']['default_profile'] = 'none'
-    elif valid_name(settings, name):
-        print("Default profile: {}".format(name))
-        settings.config['global']['default_profile'] = name
+    if not valid_name(settings, name):
+        abort()
```
mikeswoods: Nice, better (and probably more correct) than `sys.exit(1)`. Generates the proper signal too

blackfynn/cli/bf_profile.py
```diff
@@ -273,19 +302,27 @@ def show_status(bf):
 
 #Helper functions
 #=======================================================
+
+def abort():
```
mikeswoods: There is `os.abort()` as well

PR #183, opened 2019-11-18
ignoring windows prefix

JeremyLecoeur:
# Description

Windows sometimes return the file path with "\\?\" in front, which confuses the upload manager and cause the upload to never "finish" since we never update the status of the file

## Type of change

# How Has This Been Tested?

By uploading directories from both windows and mac machines


JeremyLecoeur:
the \\? is a prefix for filepaths that are long (not sure how it works though)

rlmv:
Right, but where is that value being introduced? Is it being returned to the Python client from the Agent, or is the Python client getting it when it makes sys calls to build the initial list of target files? Just curious because it might change the best place to put this conversion.

JeremyLecoeur:
It comes from the response of the agent

rlmv:
Ok 👍

PR #184, opened 2019-11-18
Fix for windows agent

JeremyLecoeur:
# Description

Fixing a typo that prevents the file path to be correctly computed for windows

## Type of change


JeremyLecoeur:
@rlmv : I had those changes locally. did not realize I forgot to add them to the final push. Good call!

PR #185, opened 2019-11-25
Get summary

JeremyLecoeur:
# Description

https://app.clickup.com/t/324u9d

In order to power a "dataset" of datasets, we need to add some functionalities in the python library  in order to make specific info about a dataset available.

## Type of change

# Checklist:

### DIFF
--- a/blackfynn/api/data.py
+++ b/blackfynn/api/data.py
@@ -68,7 +68,7 @@ class DatasetsAPI(APIBase):
 
     def owner(self, ds):
 
-        return next(iter(filter(lambda x: x.role == 'owner', self.collaborator_users(ds))))
+        return next(iter(filter(lambda x: x.role == 'owner', self.user_collaborators(ds))))
 
     def get_by_name_or_id(self, name_or_id):
         """
--- a/blackfynn/api/data.py
+++ b/blackfynn/api/data.py
@@ -38,6 +38,42 @@ class DatasetsAPI(APIBase):
         resp = self._get( self._uri('/{id}', id=id))
         return Dataset.from_dict(resp, api=self.session)
 
+    def published(self,ds):
+        id = self._get_id(ds)
+        resp = self._get( self._uri('/{id}/published', id=id))
+        return resp
+
+    def get_doi(self,ds):
+        id = self._get_id(ds)
+        resp = self._get( self._uri('/{id}/doi', id=id))
+        return resp
+
+    def get_storage(self, ds):
+        id = self._get_id(ds)
+        resp = self._get( self._uri('/{id}', id=id))
+        if "storage" in resp.keys():
+            return resp["storage"]
+        else:
+            return 0
+
+    def get_files_count(self,ds):
+        id = self._get_id(ds)
+        resp = self._get( self._uri('/{id}/packageTypeCounts', id=id))
+        file_count = 0
+        for key, value in resp.items():
+            file_count += value
+        return file_count
+
+    def get_collab_teams(self,ds):
+        id = self._get_id(ds)
+        resp = self._get( self._uri('/{id}/collaborators/teams', id=id))
+        return resp
+
+    def get_collab_users(self,ds):
+        id = self._get_id(ds)
+        resp = self._get( self._uri('/{id}/collaborators/users', id=id))
+        return resp
+
     def get_by_name_or_id(self, name_or_id):
         """
         Get Dataset by name or ID.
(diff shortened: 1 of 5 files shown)


rlmv:
It's also totally reasonable to keep some of this code out of the Python client, and implement it in the SPARC script if adding all these endpoints feels like too large of a lift. It's pretty easy to extend the from external packages. That said, I think this is all super useful information to include in the client.

blackfynn/models.py
```diff
@@ -2126,6 +2140,83 @@ def _get_method(self):
...
+class PublishInfo(object):
+
+    def __init__(self, status, latest_doi, sourceOrganizationId, sourceDatasetId, publishedDatasetId, publishedVersionCount, lastPublishedDate):
+        self.status = status
+        self.latest_doi = latest_doi
+        self.sourceOrganizationId = sourceOrganizationId
+        self.sourceDatasetId = sourceDatasetId
+        self.publishedDatasetId = publishedDatasetId
+        self.publishedVersionCount = publishedVersionCount
+        self.lastPublishedDate = lastPublishedDate
```
rlmv: What if this was just `last_published` instead of `last_published_date`?

blackfynn/api/data.py
```diff
@@ -38,6 +41,37 @@ def get(self, ds):
...
+    def get_collab_users(self,ds):
+        id = self._get_id(ds)
+        resp = self._get( self._uri('/{id}/collaborators/users', id=id))
+        return [UserCollaborator.from_dict(u) for u in resp]
+
+    def get_owner(self, ds):
+        id = self._get_id(ds)
+        resp = self._get( self._uri('/{id}/collaborators/users', id=id))
+        owner = [elt for elt in resp if elt["role"] == "owner"]
+        return [UserCollaborator.from_dict(elt) for elt in owner]
```
rlmv: Could this return single object instead of an array of owner objects? There should only be one owner.

rlmv: This method could reuse the `get_collab_users`  be something like:

```
def get_owner(self, ds):
    return next(user in self.get_collab_users(ds) if user.role == 'owner') 
```

blackfynn/models.py
```diff
@@ -726,6 +723,9 @@ def as_dict(self):
     @classmethod
     def from_dict(cls, data, *args, **kwargs):
         item = super(BaseCollection, cls).from_dict(data, *args, **kwargs)
+
+        if 'storage' in data:
+            item.storage = data['storage']
```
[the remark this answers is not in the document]
JeremyLecoeur: 👍

blackfynn/api/data.py
JeremyLecoeur: :+1

blackfynn/models.py
```diff
@@ -1929,6 +1927,21 @@ def get_graph_summary(self):
         """ Returns summary metrics about the knowledge graph """
         return self._api.concepts.get_summary(self)
 
+    def publish_info(self):
```
rlmv: This is getting pretty picky and in the weeds, but what do you think about using `published()` vs `publish_info()`? And what about removing prefixes from the `published_dataset_id` and `published_version_count` models so the variables are just `id` and `version_count`?

I'm imaging usage of it like the following:

```
>>> dataset.published().id 
>>> dataset.published().version_count
``` 
versus the existing more verbose
```
>>> dataset.published_info().published_dataset_id
>>> dataset.published_info().published_version_count
```

This feels kind of nitpicky now, but I'm imagining wanting to extend this object with more information from Discover in the future. I think this would change the `PublishInfo` object to be more of a `PublishedDataset` object. We could then add, for example:

```
>>> dataset.published().versions()
>>> dataset.published().files(path="files/test/data.dicom")
```

Eventually the top-level `Blackfynn` client could expose Discover directly, returning the same API objects
```
>>> bf.discover.datasets()[2].id
>>> bf.discover.datasets()[2].versions()
>>> bf.discover.datasets()[2].files()
```

What do you think?

blackfynn/models.py
```diff
@@ -726,6 +723,8 @@ def as_dict(self):
     @classmethod
     def from_dict(cls, data, *args, **kwargs):
         item = super(BaseCollection, cls).from_dict(data, *args, **kwargs)
+
+        item.storage = data.get('storage', 0)
```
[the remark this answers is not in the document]
JeremyLecoeur: I'm guessing the front-end relies on None, so let's have None instead of 0

PR #187, opened 2019-12-16
we can now decode a stringified boolean

JeremyLecoeur:
# Description

When the trials module create review, it passes "False" which is currently not recognized by the python code. It applies a conversion that transform any string to True.

## Github Issue

https://app.clickup.com/t/3e5m9t

## Type of change


rlmv:
Can we get a test for this, for both the true and false cases? Creating a model with a Boolean property, and creating records of that model should be sufficient.

Does the encoder need to be updated to serialize Boolean values to strings?

JeremyLecoeur:
Working on the test.

I don't think we need to do the encoder though.

CHANGELOG.md
```diff
@@ -6,9 +6,14 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
...
 
+## 3.1.1
+
+### Fixed
+- Stringified boolean are now correctly deserialized to the proper value
+
 ## 3.1.0
 
-###Added
+###A dded
```
rlmv: Typo

CHANGELOG.md
```diff
@@ -6,9 +6,14 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 Documentation at <https://developer.blackfynn.io/python/>
 
+## 3.1.1
+
+### Fixed
+- Stringified boolean are now correctly deserialized to the proper value
```
rlmv: Suggestion: what about "- Fixed a bug in the deserialization of boolean properties" ?

PR #189, opened 2019-12-19
Model service host

JeremyLecoeur:
# Description

adding the possibility to specify a different model_service_host

## Type of change


blackfynn/config.py
```diff
@@ -136,6 +136,7 @@
 
     BLACKFYNN_USE_CACHE: 0  (false) or 1  (true)  # `use_cache`
     BLACKFYNN_API_LOC                             # `api_host`
+    BLACKFYNN_MODEL_LOC                           # `model_service_host`
```
mattusifer: this won't work right? I don't think we should allow people to configure this based on environment - this is really specific to us.

JeremyLecoeur: ha ha... the one thing I wanted to remove and forgot... good catch

PR #190, opened 2020-01-24
Status log end point

JeremyLecoeur:
# Description

This PR exposes the new status-log endpoint as a function on the Dataset object.
limit and offset can be passed but are defaulted to 25 and 0 if nothing is passed.
This PR also locks the status of dataset as a read-only properties and raises an AttributeError if one attempts to change the status of a dataset

## ClickUp Issue

https://app.clickup.com/t/30cztr

## Type of change

# How Has This Been Tested?

The functionality is tested in the API itself. 
We do however have a test that checks that on creation of a dataset, we indeed create a dataset status and are able to retrieve it with all the correct info

# Checklist:

### DIFF
--- a/blackfynn/models.py
+++ b/blackfynn/models.py
@@ -1902,7 +1902,7 @@ class Dataset(BaseCollection):
         kwargs.pop('type', None)
         super(Dataset, self).__init__(name, "DataSet", **kwargs)
         self.description = description or ''
-        self.status = status
+        self._status = status
         self.automatically_process_packages = automatically_process_packages
 
         # remove things that do not apply (a bit hacky)
@@ -1912,6 +1912,15 @@ class Dataset(BaseCollection):
     @as_native_str()
     def __repr__(self):
         return u"<Dataset name='{}' id='{}'>".format(self.name, self.id)
+
+    @property
+    def status(self):
+        """Get the current status."""
+        return self._status
+
+    @status.setter
+    def status(self, value):
+        raise AttributeError('Dataset.status is read-only.')
 
     def get_topology(self):
         """ Returns the set of Models and Relationships defined for the dataset
@@ -1929,6 +1938,9 @@ class Dataset(BaseCollection):
 
     def published(self):
         return self._api.datasets.published(self.id)
+
+    def status_log(self, limit = 25, offset = 0):
+        return self._api.datasets.status_log(self.id, limit, offset)
 
     def package_count(self):
         return self._api.datasets.package_count(self.id)
@@ -2119,7 +2131,7 @@ class Dataset(BaseCollection):
             name = self.name,
             description = self.description,
             automaticallyProcessPackages = self.automatically_process_packages,
-            properties = [p.as_dict() for p in self.properties]
+            properties = [p.as_dict() for p in self.properties],
         )
 
 # ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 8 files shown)


mikeswoods:
Otherwise, looks good!

tests/test_basic.py
```diff
@@ -33,6 +33,27 @@ def test_update_dataset(client, dataset, session_id):
...
+    assert dataset_status_log.entries[0].status.id == 1
+    assert dataset_status_log.entries[0].status.name == 'NO_STATUS'
+    assert dataset_status_log.entries[0].status.display_name == 'No Status'
+
+def test_status_is_readonly(client, dataset):
+    try:
+        dataset.status = 'New Thing'
+        dataset.update()
+    except AttributeError as e:
+        assert str(e) == "Dataset.status is read-only."
```
rlmv: This actually doesn't check that an exception is always raised - just that *if* and exception is raised then it will be an `AttributeError` with this message. Pytest has a `pytest.raises(...)` context manager that is preferred for tests like this. 

https://docs.pytest.org/en/latest/assert.html#assertions-about-expected-exceptions

blackfynn/models.py
```diff
@@ -1913,6 +1913,15 @@ def __init__(self, name, description=None, status=None, automatically_process_pa
...
         return u"<Dataset name='{}' id='{}'>".format(self.name, self.id)
 
+    @property
+    def status(self):
+        """Get the current status."""
+        return self._status
+
+    @status.setter
+    def status(self, value):
+        raise AttributeError('Dataset.status is read-only.')
```
rlmv: Nice.

blackfynn/models.py
```diff
@@ -2166,6 +2178,103 @@ def from_dict(cls, data):
...
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+# StatusLogEntry
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+class StatusLogEntry(BaseNode):
+
+    def __init__(self, user, status, updated_at):
+        self.user = user
+        self.status = status
+        self.updated_at = updated_at
```
rlmv: Could this be parsed into a `datetime` object?

blackfynn/models.py
```diff
@@ -2166,6 +2178,103 @@ def from_dict(cls, data):
...
+
+
+    @classmethod
+    def from_dict(cls, data):
+        return cls(
+            user = UserStubDTO.from_dict(data.get('user')),
+            status = DatasetStatusStub.from_dict(data.get('status')),
+            updated_at = datetime.datetime.strptime(data.get('updatedAt'), "%Y-%m-%dT%H:%M:%S.%fZ"),
```
mikeswoods: If this ismeant to be an ISO8601 date, you can import
```
import dateutil.parser
```
then
```
 dateutil.parser.isoparse()
```
and not have to provide a formatting string

mikeswoods: I should mention`dateutil` is part of the standard library

rlmv: `dateutil` is actually *not* part of the standard library.

mikeswoods: oof you're right. my bad

rlmv: We could add it as a dependency, but not sure it is worth it for one endpoint. Maybe we should use `strptime` for now, and add the dependency later if we end up handling more dates?

JeremyLecoeur: we are already importing dateutil in blackfynn.models, so we would not be added a new dependency

rlmv: Oh, interesting. Something must pull that in as a transitive dependency. Should we explicitly list it in `requirements.txt` then?

JeremyLecoeur: I used JohnnyDep to solve dependency and the result is python-dateutil>=2.8.0

rlmv:
`johnnydep` is a cool tool! I didn't know that existed. I'm curious - where does the `2.8.0` come from? I just ran it myself and saw `2.6.1` as the minimum shared version between our other dependencies.

blackfynn/models.py
JeremyLecoeur: My bad, looks like I add some weirdness in my virtualenv. rerunning it from scratch gives me `2.6.1`

PR #191, opened 2020-02-03
fixing test

JeremyLecoeur:
# Description

Now that model-service is fixed, the test_related_records_pagination is still failing because of the ordering of the response. By adding a `sort()` call, we fix the issue

## Github Issue

https://app.clickup.com/t/48chfp

## Type of change

### DIFF
--- a/tests/test_concepts.py
+++ b/tests/test_concepts.py
@@ -469,7 +469,7 @@ def test_related_records_pagination(dataset):
     # Get all records
     gotten = patient1.get_related()
     assert len(gotten) == 200
-    assert [r.get("field") for r in gotten] == list(map(str, range(200)))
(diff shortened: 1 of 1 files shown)


tests/test_concepts.py
```diff
@@ -469,7 +469,7 @@ def test_related_records_pagination(dataset):
     # Get all records
     gotten = patient1.get_related()
     assert len(gotten) == 200
-    assert [r.get("field") for r in gotten] == list(map(str, range(200)))
+    assert [r.get("field") for r in gotten].sort() == list(map(str, range(200))).sort()
```
rlmv: This is technically a regression too - do you think we should fix the `model-service` endpoint to always order related objects by Neo4j node ID?

JeremyLecoeur: true. Let's kill this PR

PR #193, opened 2020-02-13
Tags setter

JeremyLecoeur:
# Description

In order to facilitate adding tags to SPARC datasets via a python script, we need to have the ability to edit the `tags` field of a Dataset. This PR add a setter on tags that checks that only a list of string can be passed.

## Type of change

# How Has This Been Tested?

Tests were added to make sure we reject strings and array of integers but accept an array of strings

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 Documentation at <https://developer.blackfynn.io/python/>
 
+## 3.4.0
+
+### Added
+- Support for Dataset.Tags
+
 ## 3.3.1
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ Documentation at <https://developer.blackfynn.io/python/>
 ## 3.4.0
 
 ### Added
-- Support for Dataset.Tags
+- Support for `Dataset.tags`
 
 ## 3.3.1
 
--- a/blackfynn/models.py
+++ b/blackfynn/models.py
@@ -1904,7 +1904,7 @@ class Dataset(BaseCollection):
         super(Dataset, self).__init__(name, "DataSet", **kwargs)
         self.description = description or ''
         self._status = status
-        self._tags = tags
+        self._tags = tags or []
         self.automatically_process_packages = automatically_process_packages
 
         # remove things that do not apply (a bit hacky)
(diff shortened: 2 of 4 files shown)


blackfynn/models.py
```diff
@@ -1919,10 +1920,22 @@ def status(self):
...
+        """Get the current tags."""
+        return self._tags
+
     @status.setter
     def status(self, value):
         raise AttributeError('Dataset.status is read-only.')
 
+    @tags.setter
+    def tags(self, value):
+        if bool(value) and isinstance(value, list) and all(isinstance(elem, str) for elem in value):
```
rlmv: Three small things:

1. In Python we typically just say `if value` instead of `if bool(value`) - they mean the same thing, and the first one is more concise.
2. But, more importantly, checking that `value` is truthy at all here will cause problems when a user tries to set `tags` to an empty list. `bool([]) == False`, but that should be allowed. I think we can remove that part of the check entirely. As long as `value` is a list and all entries are strings we should be good.
3. This should check if each tag is an instance of `string_types` instead of `str`. `string_types` is a Python 2/3 compatible shim that we need to use for full compatibility.

tests/test_basic.py
```diff
@@ -55,6 +55,17 @@ def test_status_is_readonly(client, dataset):
...
+def test_tags_is_list_of_strings_only(client, dataset):
+    with pytest.raises(AttributeError) as excinfo:
+        dataset.tags = 'New Thing'
+    assert "Dataset.tags should be a list of strings." in str(excinfo.value)
+    with pytest.raises(AttributeError) as excinfo:
+        dataset.tags = [1,2,3]
+    assert "Dataset.tags should be a list of strings." in str(excinfo.value)
+    dataset.tags = ["a","b","c"]
+    dataset.update()
+    assert dataset.tags == ["a","b","c"]
```
rlmv: Could we add one more line to this test that gets the dataset fresh from the platform and asserts that the tags are still set?

CHANGELOG.md
```diff
@@ -6,6 +6,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 Documentation at <https://developer.blackfynn.io/python/>
 
+## 3.4.0
+
+### Added
+- Support for Dataset.Tags
```
rlmv: "Dataset.Tags" should be "\`Dataset.tags\`" (lowercase, and with markdown formatting.

blackfynn/models.py
```diff
@@ -1898,12 +1898,13 @@ def __repr__(self):
...
 
 class Dataset(BaseCollection):
-    def __init__(self, name, description=None, status=None, automatically_process_packages=False, **kwargs):
+    def __init__(self, name, description=None, status=None, tags=None, automatically_process_packages=False, **kwargs):
         kwargs.pop('package_type', None)
         kwargs.pop('type', None)
         super(Dataset, self).__init__(name, "DataSet", **kwargs)
         self.description = description or ''
         self._status = status
+        self._tags = tags
```
rlmv: Should this be `self._tags = tags or []` to ensure that tags is always a list (instead of `None`)?

PR #200, opened 2020-03-02
Get package by filename

JeremyLecoeur:
# Description

Expose the newly created endpoint to get a package by a filename (needed for automatic update of SPARC metadata)

## Type of change

### DIFF
--- a/blackfynn/api/data.py
+++ b/blackfynn/api/data.py
@@ -14,6 +14,7 @@ from blackfynn.models import (
     BaseDataNode,
     Collection,
     Dataset,
+    DataPackage,
     File,
     Organization,
     Tabular,
@@ -22,7 +23,6 @@ from blackfynn.models import (
     User,
     StatusLogResponse,
     StatusLogEntry,
-    PackagesResponse,
     PublishInfo,
     UserCollaborator,
     TeamCollaborator,
@@ -77,7 +77,7 @@ class DatasetsAPI(APIBase):
     def get_packages_by_filename(self,ds,filename):
         id = self._get_id(ds)
         resp = self._get( self._uri('/{id}/packages?filename={filename}', id=id, filename=filename))
-        return PackagesResponse.from_dict(resp)
+        return [DataPackage.from_dict(p) for p in resp.get('packages')]
 
     def owner(self, ds):
         return next(iter(filter(lambda x: x.role == 'owner', self.user_collaborators(ds))))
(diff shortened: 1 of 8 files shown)


tests/test_io.py
```diff
@@ -30,6 +32,22 @@ def test_upload_legacy_to_dataset(dataset):
...
+    ([FILE4], 1),          # Single file
+])
+def test_get_by_filename(dataset, upload_args, n_files):
+    """
+    Note: ETL will fail since destination will likely be removed
+          before being processed.
+    """
+    dataset.upload(*upload_args)
+    dataset.update()
+    time.sleep(1)
```
rlmv: Depending on any particular sleep interval will probably cause intermittent test failures down the line. Could this wait until the package is ready?

rlmv:
Instead of returning a `PackagesReponse` object with a nested `packages` key that the unser needs to unpacke could `get_packages_by_filename` just return a list of packages instead? That would both remove an extra class from the code, and simplify interacting with this endpoint.

tests/test_io.py
JeremyLecoeur: done

tests/test_io.py
```diff
@@ -30,6 +32,27 @@ def test_upload_legacy_to_dataset(dataset):
...
+    """
+    dataset.upload(*upload_args)
+    dataset.update()
+    for i in range(5):
+        pp = dataset.get_items_by_name("test-78f3ea50-b")
+        if (pp[0].state == 'READY'):
+            break
+        else:
+            time.sleep(0.5)
+    p = dataset.get_packages_by_filename('test-78f3ea50-b')
```
rlmv: Interesting, the endpoint doesn't let you include the extension in the filename when making requests? Is that a bug?

blackfynn/api/data.py
```diff
@@ -73,8 +74,12 @@ def user_collaborators(self,ds):
         resp = self._get( self._uri('/{id}/collaborators/users', id=id))
         return [UserCollaborator.from_dict(u) for u in resp]
 
-    def owner(self, ds):
+    def get_packages_by_filename(self,ds,filename):
+        id = self._get_id(ds)
+        resp = self._get( self._uri('/{id}/packages?filename={filename}', id=id, filename=filename))
```
mikeswoods: Does filename refer to the basename of the file (sans extension) or the whole thing? Can you do something like `filename=foo.txt`?

JeremyLecoeur: The underlying endpoint in bf-api only takes the basename of the file and returns all matches and lets the user deal with it

PR #201, opened 2020-03-05
Add api to package

JeremyLecoeur:
# Description

When construction the `Package` objects in the `get_package_by_filename` function, I omitted to pass the session. Currently, these objects do not have access to any api functionality. This PR fixes the issue

## Type of change

# Checklist:

### DIFF
--- a/blackfynn/models.py
+++ b/blackfynn/models.py
@@ -853,8 +853,8 @@ class DataPackage(BaseDataNode):
 
     @classmethod
     def from_dict(cls, data, *args, **kwargs):
+        data['content']['id'] = data['content']['nodeId']
         item = super(DataPackage, cls).from_dict(data, *args, **kwargs)
-
         # parse objects
         objects = data.get('objects', None)
         if objects is not None:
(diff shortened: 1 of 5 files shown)


rlmv:
How about a little test? Could just add a line to the existing `get_packages_by_filename` test checking that you can get the files from it?

blackfynn/models.py
```diff
@@ -853,8 +853,8 @@ def as_dict(self):
 
     @classmethod
     def from_dict(cls, data, *args, **kwargs):
+        data['content']['id'] = data['content']['nodeId']
```
rlmv: Is this required to instantiate every package object, or just ones from this endpoint? If the latter, might be more appropriate to just update the response from that endpoint. However the tests do seem to be passing.

JeremyLecoeur: I was wondering exactly that: where does it make more sense to do this operation. The issue seems to be that some endpoints return a package with Id being the intID, some with Id being the nodeID. But in every case nodeID is filled with the proper info. Any preference?

rlmv: It could go either way, but in general I think it is better to explicitly special-case the special cases. If all endpoints do return a `nodeId` then if may not matter though.

PR #205, opened 2020-03-16
fixing ordering in test

JeremyLecoeur:
# Description

Looks like the same test that was fixed is now failing. Since it looks like we can not rely on the query to always sort the objects in the response consistently, I propose to sort the results before comparison.

## Type of change

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ install:
   - pip install --upgrade -r requirements-test.txt
 
 
-script: pytest
+script: pytest -s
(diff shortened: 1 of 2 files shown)


rlmv:
Why can't we rely on the code to sort correctly? Is the endpoint sorting by `createdAt` timestamp and these objects all have the same timestamp?

JeremyLecoeur:
That's indeed what I saw when looking at the objects

rlmv:
Hmm. I'm 95% sure we can still resolve this in `model-service`. Let me try and throw together a quick PR.

PR #206, opened 2020-04-01
feat(batch create linked properties)

matthewtgilbride:
# Description

Add a method passing through to the endpoint in model-service

Clickup: #5ee8px

## Type of change

# Checklist:

### DIFF
--- a/blackfynn/api/concepts.py
+++ b/blackfynn/api/concepts.py
@@ -445,6 +445,41 @@ class RecordsAPI(ModelsAPIBase):
         return LinkedModelValue.from_dict(resp, source_model=concept,
             target_model=target, link_type=link_type)
 
+    def create_link_batch(self, dataset, concept, instance, payload):
+        dataset_id = self._get_id(dataset)
+        concept_id = self._get_id(concept)
+        instance_id = self._get_id(instance)
+
+        json = {
+            "data": payload
+        }
+
+        resp = self._post(
+            self._uri(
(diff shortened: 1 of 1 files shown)


matthewtgilbride:
@bo @mikeswoods @JeremyLecoeur this is my first PR for the python client.  I don't see any relevant place to add a unit test (did a usage search on `class RecordsAPI`).  Can I run this locally to test against the non-prod model-service?

rlmv:
@matthewtgilbride We don't have any direct tests of the API code because that is all our private API. Instead we test the abstraction layer in `models.py`. However, since this is going to be used by an internal script for now maybe we don't need to add a method to `models.py`

The tests will use your default API key/secret configured in your `~/.blackfynn/config.ini` file.

blackfynn/api/concepts.py
```diff
@@ -445,6 +445,46 @@ def create_link(self, dataset, concept, instance, payload):
...
+        )
+
+        results = resp["data"]
+
+        return results
+        # TODO: LinkedModelValue requires a fully instantiated target model
+        # it's quite inefficient to retrieve individual models in a loop via `concepts.get`,
+        # but we don't have a better way to retrieve them at this time.
+        # For the sake of speed, we just return the raw JSON for now
+        # linked_model_values = []
```
matthewtgilbride, line 474: After speaking with @rlmv, we realized that the only way to return `LinkedModelValue`s is to make inefficient calls to get full target model instances.  Since the motivation at this point is to unblock @jwagenaar , we just return the raw JSON from the call to API.

PR #213, opened 2020-07-10
Deprecate `default` model property field

rlmv:
# Description

The `required` field should be used instead of `default` to mark a
property as required for record creation. This PR updates the client to
set `default` based on the value of `required`, and set `required=True`
if the property is the title of the model.

Ref: https://app.clickup.com/t/u2ya7

## Type of change

# How Has This Been Tested?

Added unit tests. Created model with the new code and verified that the
frontend no longer requires properties with `required=False`.

# Checklist:


clohr:
Task linked: [CU-u2ya7 Creating models with Python client makes each property required](https://app.clickup.com/t/u2ya7)

PR #215, opened 2020-07-10
Rollback model creation when properties are invalid

rlmv:
# Description

When creating a model with properties, if creating properties fails the
client swallows the error and logs an error message, leaving a model
with no properties. This is not a great experience, as the user believes
the model was created successfully; however future requests to create
records from the model will fail because there are no properties.

This PR updates the client to delete the newly created mode in the case
that properties cannot be created and raise an exception. This should
make it easier for users to retry without have to manually delete the
model or update properties on an existing model.

## Type of change

# How Has This Been Tested?

Updated unit test.

# Checklist:


clohr:
Task linked: [CU-u2ya7 Creating models with Python client makes each property required](https://app.clickup.com/t/u2ya7)

